PR #2000, opened 2020-11-18
Transaction comment

### LINKED ISSUE #1985: Add comment to all transactions
Add a comment which displays why a transaction occurred. Update all endpoints to add a comment.

hschu12:
Added transaction comment.

If an endpoint is needed to just get transactions for a project let me know,

Fixes #1985


DanThrane (commented):
I think it would be better if we expand the comment with slightly more details about the transaction. It could just be a text field which states something like "Transferred XXX DKK to project YYY"

DanThrane (approved):

Issue #1932, opened 2020-10-21
Search By Tag (apps) should sort results 


hschu12:
Using the See all function on app store makes it hard to get a overview of the apps since they are randomly shuffled (ordered by ID?)

Sort results alphabetically or by tool.

PR #2032, opened 2020-12-03
App issues

### LINKED ISSUE #1931: Search by tags does not filter on backend
The tags cards on app-store is getting the 50 first apps with a certain tag, and then filters on them. This results in some cards being quite empty since they filter out many tools and therefore do not include more essential apps futher down the result list.
Upgrade appstores searchByTag to take a list of predetermined tools and filter them in the search allowing non tool apps to become part of the results.

hschu12:
Fixes #1931 and #1932


frontend-web/webclient/app/Applications/Overview.tsx
```diff
@@ -57,6 +57,17 @@ function Applications(props: ApplicationsProps): JSX.Element {
...
+        "bedtools",
+        "cellranger",
+        "homer",
+        "kallisto",
+        "macs2",
+        "salmon",
+        "samtools",
+        "seqtk"
+    ]
+
```
Tiggles, line 70: Just to make sure, these are lowercased in the backend as well, right?

frontend-web/webclient/app/Applications/Overview.tsx
```diff
@@ -103,7 +114,7 @@ function Applications(props: ApplicationsProps): JSX.Element {
             />
 
             {featuredTags.map(tag =>
-                <TagGrid key={tag} favorites={favPairs} tag={tag} omit={defaultTools} rows={1} columns={7}
```
Tiggles, line 106: If the backend excludes them, then the `omit` field shouldn't be necessary, should it?

frontend-web/webclient/app/Applications/Redux/BrowseActions.tsx
```diff
@@ -26,13 +26,14 @@ export const receivePage = (page: Page<FullAppInfo>): ReceiveApp => ({
...
 
-export const fetchByTag = async (tag: string, itemsPerPage: number, page: number): Promise<ReceiveApp> => ({
+export const fetchByTag = async (tag: string, excludeTools: string[], itemsPerPage: number, page: number): Promise<ReceiveApp> => ({
     type: Tag.RECEIVE_APP,
     payload: await unwrapCall(
         Client.get<Page<FullAppInfo>>(buildQueryString(
             "/hpc/apps/searchTags",
             {
                 query: tag,
+                excludeTools: excludeTools,
```
Tiggles, line 36:
```suggestion
                excludeTools,
```

frontend-web/webclient/app/Applications/Overview.tsx
hschu12: Yes they are, I found out that e.g. cellrangers tool name was cellranger while the tag name was Cell Ranger so there was a need for multiple lists. Most of the others worked by just lowercasing them on the backend.

Tiggles: Is that a good approach?

hschu12: We general need a better approach than hard coding the tools and tags needed. Should perhaps be configurable from config file to the backend. But this might something that is moved to when we also redesign app tags.
This would also be a major rewrite of frontend code if it is the backend that determines which tools and tags will be shown.

Tiggles: It's not so much the hardcoding as much as the uppercase/lowercase aspect of it for me. The inconsistencies between calls makes coding/debugging a bit more problematic.

Tiggles: But this is a rewrite issue, more than a current iteration issue

backend/app-store-service/api/src/main/kotlin/api/AppStore.kt
```diff
@@ -104,6 +104,7 @@ data class SetPublicRequest(
 
 data class TagSearchRequest(
     val query: String,
+    val excludeTools: List<String>? = emptyList(),
```
DanThrane, line 107:
```suggestion
    val excludeTools: String? = null,
```

backend/app-store-service/src/main/kotlin/app/store/rpc/AppSearchController.kt
```diff
@@ -14,7 +14,16 @@ class AppSearchController (
     override fun configure(rpcServer: RpcServer): Unit = with(rpcServer) {
 
         implement(AppStore.searchTags) {
-            ok(searchService.searchByTags(ctx.securityPrincipal, ctx.project, request.tags, request.normalize()))
+            val normalizedExcludeList = request.excludeTools?.firstOrNull()?.split(",") ?: emptyList()
```
DanThrane, line 17:
```suggestion
            val normalizedExcludeList = request.excludeTools?.split(",") ?: emptyList()
```

PR #2045, opened 2020-12-09
Support jira intergration

### LINKED ISSUE #1999: Integrate Support tickets with JIRA
Currently we only recieve support tickets in slack. We need to integrate this with our JIRA platform.
A perfect solution would be an in-platform solution where we integrate JIRA directly into our support field. 

For now we might just be able to send an email from the backend to our support mailbox with the original sender as email sender. Since JIRA already polls from support@escience.sdu.dk this should fix the problem. 

Bug reports should still only send to slack, or be integrated with github (What do you think @DanThrane ).

(Issue number https://www.youtube.com/watch?v=weUhBGA8mxo )

hschu12:
Adds subject field to support tickets and forward tickets to support@escience.sdu.dk as well as slack.

This have not been tested since localhost gets 404 on send slack and dev does not work.
If works will fix #1999

### DIFF
--- a/backend/mail-service/api/src/main/kotlin/MailDescriptions.kt
+++ b/backend/mail-service/api/src/main/kotlin/MailDescriptions.kt
@@ -22,8 +22,32 @@ data class SendBulkRequest(
     val messages: List<SendRequest>
 )
 
+data class SendSupportEmailRequest(
+    val fromEmail: String,
+    val subject: String,
+    val message: String
+)
+
 object MailDescriptions : CallDescriptionContainer("mail") {
     val baseContext = "/api/mail"
+
+    val sendSupport = call<SendSupportEmailRequest, Unit, CommonErrorMessage>("sendSupport") {
+        auth {
+            roles = setOf(Role.SERVICE)
+            access = AccessRight.READ_WRITE
+        }
+
+        http {
+            method = HttpMethod.Post
+
+            path {
+                using(baseContext)
+                +"support"
+            }
+
+            body { bindEntireRequestFromBody() }
+        }
+    }
 
     val send = call<SendRequest, Unit, CommonErrorMessage>("send") {
         auth {
--- a/backend/mail-service/api/src/main/kotlin/MailDescriptions.kt
+++ b/backend/mail-service/api/src/main/kotlin/MailDescriptions.kt
@@ -28,10 +28,12 @@ data class SendSupportEmailRequest(
     val message: String
 )
 
+typealias SendSupportEmailResponse = Unit
+
 object MailDescriptions : CallDescriptionContainer("mail") {
     val baseContext = "/api/mail"
 
-    val sendSupport = call<SendSupportEmailRequest, Unit, CommonErrorMessage>("sendSupport") {
+    val sendSupport = call<SendSupportEmailRequest, SendSupportEmailResponse, CommonErrorMessage>("sendSupport") {
         auth {
             roles = setOf(Role.SERVICE)
             access = AccessRight.READ_WRITE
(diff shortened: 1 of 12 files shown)


backend/mail-service/api/src/main/kotlin/MailDescriptions.kt
```diff
@@ -22,9 +22,33 @@ data class SendBulkRequest(
...
+data class SendSupportEmailRequest(
+    val fromEmail: String,
+    val subject: String,
+    val message: String
+)
+
 object MailDescriptions : CallDescriptionContainer("mail") {
     val baseContext = "/api/mail"
 
+    val sendSupport = call<SendSupportEmailRequest, Unit, CommonErrorMessage>("sendSupport") {
```
Tiggles, line 34:
```suggestion
    val sendSupport = call<SendSupportEmailRequest, SendSupportEmailResponse, CommonErrorMessage>("sendSupport") {
```

Also add `typealias SendSupportEmailResponse = Unit`.

backend/mail-service/src/main/kotlin/mail/services/MailService.kt
```diff
@@ -106,6 +106,63 @@ class MailService(
...
+        fromEmail: String,
+        subject: String,
+        text: String
+    ) {
+        val recipientAddress = InternetAddress("support@escience.sdu.dk")
+
+        // Setup mail server
+        val properties = System.getProperties()
+        properties.setProperty("mail.smtp.host", "localhost")
+        properties.setProperty("mail.smtp.port", "25")
```
Tiggles, line 134: This is done in both this function and `send`. Couldn't this be done on service initialization or is it a per request thing?

backend/support-service/src/test/kotlin/support/http/SupportTest.kt
```diff
@@ -35,7 +35,7 @@ class SupportTest {
                     method = HttpMethod.Post,
                     path = "/api/support/ticket",
                     user = TestUsers.user,
-                    request = CreateTicketRequest("This is message")
+                    request = CreateTicketRequest("subbject", "This is message")
```
Tiggles, line 38:
```suggestion
                    request = CreateTicketRequest("subject", "This is message")
```

frontend-web/webclient/app/ui-components/SupportBox.tsx
```diff
@@ -42,11 +43,13 @@ export default function Support(): JSX.Element {
     async function onSubmit(event: React.FormEvent): Promise<void> {
         event.preventDefault();
         const text = textArea.current?.value ?? "";
+        const title = titleArea.current?.value ?? "";
         if (text.trim()) {
             try {
                 setLoading(true);
-                await Client.post("/support/ticket", {message: `${type}: ${text}`});
+                await Client.post("/support/ticket", {subject: `${title}`, message: `${type}: ${text}`});
```
Tiggles, line 50:
```suggestion
                await Client.post("/support/ticket", {subject: title, message: `${type}: ${text}`});
```

Tiggles (commented):
Looks fine. Is the subject field added necessary?

hschu12:
The subject field is mostly for Jira. Here each ticket gets the name of the subject from the email. But no it is not strictly needed. It can be removed and we just add a title ourself.

DanThrane (approved):
Looks good to me. @Tiggles Should we remove the subject and just default to something else?

Tiggles:
> Looks good to me. @Tiggles Should we remove the subject and just default to something else?

I think it's fine to have. Let's keep it for now, and remove if end-users aren't using it properly.

PR #2083, opened 2021-01-06
changed SQL statement to fix #2062. 

### LINKED ISSUE #2062: If app is not public, this affects all other version
If the latest release of an app is not public, then users are not able to access all previous releases which are set as public. This is an issue in case we upload software releases for development purpose.

hschu12:
Uses Destinct and order by instead of max(createdBy). Tested locally


hschu12:
Dont ask me why so much have been removed from package lock - have updated npm and run npm install

Tiggles:
> Dont ask me why so much have been removed from package lock - have updated npm and run npm install

I think this might be related to different versions of Node/NPM. My version seems to have a different approach.

DanThrane (approved):

PR #2121, opened 2021-01-15
Attempt at product update to backend

Tiggles:


backend/app-kubernetes-service/api/src/main/kotlin/api/KubernetesLicense.kt
```diff
@@ -21,7 +21,13 @@ data class KubernetesLicense(
...
     val port: Int,
     val tags: List<String>,
-    val license: String?
+    val license: String?,
+    val category: ProductCategoryId,
+    val pricePerUnit: Long,
+    val description: String,
+    val availability: ProductAvailability,
+    val priority: Int,
+    val paymentModel: PaymentModel = PaymentModel.PER_ACTIVATION,
```
Tiggles, line 30: Should some of these use default values? Should paymentModel default be removed?

DanThrane: I would try to mirror what the `Product`s already do in terms of default values.

DanThrane (approved):

PR #2122, opened 2021-01-15
Activity stream for job starts

### LINKED ISSUE #2065: [scheduling] Reimplement activity stream for job starts
The API has changed as a result the activity stream needs to be updated to support the new flow.

xiroV:
Reimplements the activity stream for job starts, according to the new scheduling backend, and fixes a few minor issues:

 - When multiple files were used in an app, the file activity stream would only show one of the files. This have been solved by creating multiple activities (one for each file used in the app). Hence, each file will be shown as a separate activity.
 - When a file was used in an app, this was not shown in the file properties of the file. This have been solved by creating an activity if the file path matches a file which was used in an app, or if the file path is a child of a file/directory which was used in an app.

fixes #2065


backend/activity-service/src/main/kotlin/activity/services/ActivityEventElasticDao.kt
```diff
@@ -542,75 +545,54 @@ class ActivityEventElasticDao(private val client: RestHighLevelClient) {
...
-            object ApplicationStart : CallWithActivity<StartJobRequest>(
+           object ApplicationStart : CallWithActivity<JobsCreateRequest>(
                 ActivityEventType.usedInApp,
-                JobDescriptions.start,
+                Jobs.create,
                 jacksonTypeRef(),
                 listOf(
-                    "mounts.source",
-                    "parameters.*.source"
+                    "resources.path"
```
DanThrane, line 553: Missing files from parameters.

backend/activity-service/src/main/kotlin/activity/services/ActivityEventElasticDao.kt
```diff
@@ -542,75 +545,54 @@ class ActivityEventElasticDao(private val client: RestHighLevelClient) {
...
-                                        doc.requestJson.application.version
-                                    )
-                                )
-                            }
-                        }
-                        doc.requestJson.parameters.forEach { (t, u) ->
-                            if (t == "directory") {
-                                val path = checkSource(u.toString(), normalizedFilePath)
+                        doc.requestJson.items.forEach { item ->
+                            item.resources?.forEach { resource ->
```
DanThrane, line 586: Missing files passed in parameters

DanThrane (changes requested):
Missing some files, otherwise this looks good to me.

DanThrane (approved):

PR #2168, opened 2021-02-05
Hide license servers from grant application editor

### LINKED ISSUE #2107: [scheduling] Hide some products from grant application editor
This is needed to support "bring your own license server" which we are currently doing in production.

Steps to fix this:

1. Introduce a `hiddenInGrantApplications` attribute to products
2. Modify the grant application editor to only list products which does not set this to true (front- or backend, doesn't matter significantly for the moment)
3. Modify the SKUs page to only show products which doesn't have this property set to true
4. Modify the app license server editor to set this property (will require some backend tweaks)

xiroV:
fixes #2107


Tiggles (approved):
FE seems fine

backend/accounting-service/src/main/kotlin/services/Products.kt
```diff
@@ -219,7 +222,7 @@ class ProductService(
                     {
                         setParameter("provider", provider)
                     },
-                    "select * from products where provider = :provider order by priority, id"
+                    "select * from products where provider = :provider and hidden_in_grant_applications is false order by priority, id"
```
DanThrane, line 225: This almost certainly breaks parts of the backend. If not now it will later. This is supposed to be hidden in grant applications, not hidden everywhere.

backend/accounting-service/src/main/kotlin/services/Products.kt
```diff
@@ -270,7 +273,7 @@ class ProductService(
                     },
                     """
                         from products
-                        where provider = :provider AND area = :area
+                        where provider = :provider and area = :area and hidden_in_grant_applications is false
```
DanThrane, line 276: Same point as above.

backend/accounting-service/src/main/kotlin/services/Products.kt
```diff
@@ -476,6 +481,7 @@ class ProductService(
                         getField(ProductTable.provider)
                     ),
                     getField(ProductTable.description),
+                    false,
```
DanThrane, line 484: Any reason we don't use the value from the table here?

backend/accounting-service/src/main/kotlin/services/Products.kt
```diff
@@ -460,6 +464,7 @@ class ProductService(
...
```
DanThrane, line 467: Any reason we don't use the value from the table here?

backend/accounting-service/src/main/kotlin/services/Products.kt
```diff
@@ -440,6 +443,7 @@ class ProductService(
...
```
DanThrane, line 446: Any reason we don't use the value from the table here?

DanThrane (changes requested):

DanThrane (approved):

PR #2199, opened 2021-02-11
Data extraction

hschu12:
Not at all pretty code, but works for the current period for DeiC Report. Next will mosly likely need rewrite and a change of data format.


DanThrane (approved):
Looks good to me assuming that we make improvements when the DeIC format is clarified.

PR #2224, opened 2021-02-12
Make Runs UI more consistent with Files UI

### LINKED ISSUE #2196: [scheduling] Tweak styling of 'Runs' page to be more consistent with the file table
The new component is using some slightly different styling which is not super consistent with what we already had. Let us try to make it a bit more consistent:

- Spacing is different
- We don't use a border (this is inconsistent in many places, let us pick one)
- Styling of the confirm button when placed in the sidebar (i.e. delete in this case) looks bad

xiroV:
- Added borders to job elements,
 - added header,
 - decreased font size and centered text on Cancel button.

fixes #2196


DanThrane (approved):
Code looks good to me. I guess we can tweak the styling further if needed.

PR #2231, opened 2021-02-16
attempt to fix wayf user creation issue

hschu12:
We had an issue when creating a WAYF user. Perhaps due to retires there was two users created before we could check, resulting in null value (singleOrNull). This fix makes a check in insert function if the user is from WAYF to check on WAYFID instead of our user generated ID this should make the situation more unlikely.

### DIFF
--- a/backend/ucloud-data-extraction-service/src/main/kotlin/services/UserActivityReport.kt
+++ b/backend/ucloud-data-extraction-service/src/main/kotlin/services/UserActivityReport.kt
@@ -1,0 +1,4 @@
+package dk.sdu.cloud.ucloud.data.extraction.services
+
+class UserActivityReport {
+}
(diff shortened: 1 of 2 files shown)


backend/ucloud-data-extraction-service/src/main/kotlin/services/UserActivityReport.kt
```diff
@@ -0,0 +1,4 @@
+package dk.sdu.cloud.ucloud.data.extraction.services
+
+class UserActivityReport {
+}
```
DanThrane, line 4: Delete

DanThrane (commented):
Looks good to me. You can delete the extra files which I guess wasn't supposed to be committed.

Issue #2274, opened 2021-03-01
[scheduling] Job status page does not update when the job is terminated due to expiration


Tiggles:
Reported by @claudiopica

xiroV:
Just a few notes on my findings related to this issue.

**Steps to reproduce**
Open 4-5 new tabs, and start a 1-minute application in each of them. Don't leave the JobView page. When the jobs expire, a few of them will usually not transition to the "Job expired" page. If they all transition correctly, repeat.

It could be a coincidence, but it seems to happen more often when the browser tab is new, and you don't leave the JobView page (or it could be when multiple JobView pages are open at the same time?).

The transition to the "Job expired" page possibly happens due to the following error, which was found the websocket stream (`/api/jobs`):

```
stdout: (node:6) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
[1m[38;2;0;143;191mINFO  [0m[1mcode-server development
[1m[38;2;0;143;191mINFO  [0mAdditional documentation: http://github.com/codercom/code-server
[1m[38;2;0;143;191mINFO  [0mInitializing[38;2;140;140;140m {"data-dir":"/home/coder/.local/share/code-server","extensions-dir":"/home/coder/.local/share/code-server/extensions","working-dir":"/work","log-dir":"/home/coder/.cache/code-server/logs/20210322102949595"}
[1m[38;2;0;143;191mINFO  [0mStarting shared process [1/5]...
[1m[38;2;0;143;191mINFO  [0mStarting webserver...[38;2;140;140;140m {"host":"0.0.0.0","port":8443}
[1m[38;2;255;157;0mWARN  [0mNo certificate specified. [1mThis could be insecure.
[1m[38;2;255;157;0mWARN  [0mDocumentation on securing your setup: https://github.com/codercom/code-server/blob/master/doc/security/ssl.md
[1m[38;2;255;157;0mWARN  [0mLaunched without authentication.
[1m[38;2;0;143;191mINFO  [0m 
[1m[38;2;0;143;191mINFO  [0mStarted (click the link below to open):
[1m[38;2;0;143;191mINFO  [0mhttp://localhost:8443/
[1m[38;2;0;143;191mINFO  [0m 
[...]

PR #2322, opened 2021-03-19
Fix frontend crash when job expires

xiroV:
Seemed to be caused by an uncaught type error, but only happened some times, when on the JobView page while the job expired.

Ref. #2274 , but I'm not sure it's a fix.

### DIFF
--- a/frontend-web/webclient/app/Applications/Jobs/View.tsx
+++ b/frontend-web/webclient/app/Applications/Jobs/View.tsx
@@ -832,12 +832,12 @@ const RunningContent: React.FunctionComponent<{
     const appInvocation = job.specification.resolvedApplication!.invocation;
     const backendType = appInvocation.tool.tool!.description.backend;
     const support = job.specification.resolvedSupport;
-    const supportsExtension = support === null ? false :
-        (backendType === "DOCKER" && support!.docker.timeExtension) ||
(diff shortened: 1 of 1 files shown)


frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -1140,16 +1140,16 @@ const RunningButtonGroup: React.FunctionComponent<{
...
 
     const appType = appInvocation.applicationType;
-    const supportsInterface =
-        (appType === "WEB" && backendType === "DOCKER" && support.docker.web) ||
-        (appType === "VNC" && backendType === "DOCKER" && support.docker.vnc) ||
-        (appType === "VNC" && backendType === "VIRTUAL_MACHINE" && support.virtualMachine.vnc);
+    const supportsInterface = support === null ? false :
+        (appType === "WEB" && backendType === "DOCKER" && support!.docker.web) ||
+        (appType === "VNC" && backendType === "DOCKER" && support!.docker.vnc) ||
+        (appType === "VNC" && backendType === "VIRTUAL_MACHINE" && support!.virtualMachine.vnc);
```
Tiggles, line 1152: Use null-safe navigator instead (`.?`).

DanThrane (approved):

PR #2379, opened 2021-04-21
Enforce lower-case ingresses

### LINKED ISSUE #2350: Make sure that all public links are transformed to lower case before passing it to the provider
It sounds like capitalization is causing issues in public links. This could be specific to the UCloud/Compute provider. It might be better to simply transform to lowercase before passing it to the provider.

xiroV:
Save ingresses/public links as lower-case on creation.

fixes #2350

### DIFF
--- a/backend/app-orchestrator-service/src/jvmMain/kotlin/services/IngressDao.kt
+++ b/backend/app-orchestrator-service/src/jvmMain/kotlin/services/IngressDao.kt
@@ -214,7 +214,7 @@ class IngressDao(
             Ingress(
                 it.getField(IngressTable.id),
                 IngressSpecification(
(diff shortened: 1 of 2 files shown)


DanThrane (changes requested):
This should also normalize the ingresses that are being returned. Just in case a provider is returning one which isn't being lowercased.

xiroV:
I've updated the `receive` and `browse` end-points to also lower-case the domains as requested. However,

 1. It doesn't look like we are fetching the ingresses from the providers (but from our local database) as far as I can see. For backwards compatibility I think it makes sense to lowercase them anyways though.
 2. I'm not sure this is a efficient way to do it, so please let me know if there's a better option.

PR #2385, opened 2021-04-23
Removing licenses (and public ip) from low funds check

### LINKED ISSUE #2374: Turn off low balance warnings for software licenses

hschu12:
Added config and changes to SQL. Fixes #2374


DanThrane (changes requested):
This could be changed such that you join with the product table and look at the type. There is no reason to hardcode the existing products into this.

DanThrane (approved):

PR #2492, opened 2021-06-15
Storage uploads

Tiggles:
Temporary pausable uploads solution. Fix tests.


DanThrane (approved):
Looks good to me. We might need to change the instructions a bit before this goes into production but we can take a look at this later.

PR #2502, opened 2021-06-21
Issue 2340 pagev2 issue

Tiggles:
Fixes PageV2 duplicate for Browse.


frontend-web/webclient/app/Applications/Jobs/Browse.tsx
```diff
@@ -89,12 +89,11 @@ export const Browse: React.FunctionComponent = () => {
     useRefreshFunction(refresh);
 
     useLoading(jobs.loading);
-    const projectId = useProjectId();
 
     useEffect(() => {
         fetchJobs(UCloud.compute.jobs.browse({itemsPerPage, ...flags, ...filters, sortBy}));
         setInfScrollId(id => id + 1);
-    }, [projectId, filters, sortBy]);
+    }, [filters, sortBy]);
```
DanThrane, line 96: Won't this break auto-reloading of pages when changing to a new project?

Tiggles: It fixed this exact issue, that only occurs for me when changing projects on the run page.

The header already loads new info on project change, which might be a side-effect causing this to work properly, but I'm not entirely sure.

DanThrane: Right, I think you are probably right.

DanThrane (approved):

PR #2504, opened 2021-06-22
Storage uploads

Tiggles:
Improvement to resumeable uploads. Added button to resume uploads, when paused. Also allow to resume uploads if paused more than once.


frontend-web/webclient/app/Files/Uploader.tsx
```diff
@@ -80,6 +80,36 @@ async function processUpload(upload: Upload) {
...
+            await sendChunk(await reader.readChunk(maxChunkSize));
+
+            const expiration = new Date().getTime() + FOURTY_EIGHT_HOURS_IN_MILLIS;
+            localStorage.setItem(
+                createLocalStorageUploadKey(fullFilePath),
+                JSON.stringify({chunk: reader.offset, size: upload.fileSizeInBytes, response: strategy!, expiration} as LocalStorageFileUploadInfo)
+            );
+        }
+
+        console.log(upload);
```
DanThrane, line 104:
```suggestion
```

frontend-web/webclient/app/Files/Uploader.tsx
```diff
@@ -80,6 +80,36 @@ async function processUpload(upload: Upload) {
...
+    fullFilePath: string
+) {
+    return async () => {
+        while (!reader.isEof() && !upload.terminationRequested) {
+            await sendChunk(await reader.readChunk(maxChunkSize));
+
+            const expiration = new Date().getTime() + FOURTY_EIGHT_HOURS_IN_MILLIS;
+            localStorage.setItem(
+                createLocalStorageUploadKey(fullFilePath),
+                JSON.stringify({chunk: reader.offset, size: upload.fileSizeInBytes, response: strategy!, expiration} as LocalStorageFileUploadInfo)
```
DanThrane, line 100: It might be better if we rename `chunk` to `offset` since `chunk` might indicate that we are counting the number of chunks we have sent as opposed to the number of bytes (which I am pretty sure is what `reader.offset` reports).

I also think it would make more sense to rename `response` to `strategy`.

DanThrane (changes requested):
I think we can make a few minor adjustments.

DanThrane (approved):

PR #2514, opened 2021-07-01
update to mail service

hschu12:


backend/mail-service/src/jvmMain/kotlin/mail/services/MailService.kt
```diff
@@ -134,17 +134,17 @@ class MailService(
...
     ) {
         val recipientAddress = InternetAddress("support@escience.sdu.dk")
 
         try {
             val message = MimeMessage(session)
-            message.setFrom(InternetAddress(fromEmail))
+            message.setFrom("ticketsystem@escience.sdu.dk")
             message.addRecipient(Message.RecipientType.TO, recipientAddress)
-            message.subject = subject
+            message.subject = "$userEmail-|-"+subject
```
Tiggles, line 147: The mix and match of string interpolation and addition of strings is not very pretty. Also, formatting is not properly done for this.

Tiggles (changes requested):
One formatting issue

Tiggles (approved):

PR #2549, opened 2021-07-22
Syncthing (master)

### LINKED ISSUE #2506: [sync] Synchronization feature for current master branch
I've ported the synchronization (Syncthing) feature from the subbranch of `storage` ([syncthing-integration](https://github.com/SDU-eScience/UCloud/tree/syncthing-integration)) to a subbranch of `master` ([syncthing-master](https://github.com/SDU-eScience/UCloud/tree/syncthing-master)) so that we can begin testing the synchronization-feature before the storage-epic (#2201) is complete.

In the [syncthing-integration](https://github.com/SDU-eScience/UCloud/tree/syncthing-integration) branch, the feature is implemented as a part of the `file-ucloud-service`, while on [syncthing-master](https://github.com/SDU-eScience/UCloud/tree/syncthing-master) it implemented in: `storage-service`.

- [x] Permissions should be fixed such that if a user have read access (only) to a folder, the `SynchronizationType` should be set to `SEND_ONLY` when the folder is added to Syncthing, `SEND_RECEIVE` otherwise.
 - [x] Only admins should have access to this feature, such that we can test it in production, without users having access.
 - [x] It should only be possible to add folders to syncthing with a limited number of files (1.000.000).
 - [x] In case of multiple Syncthing instances configured on UCloud, the instance with the least total size of data/files attached should be chosen.

xiroV:
fixes #2506


backend/storage-service/api/src/commonMain/kotlin/FileSynchronizationDescriptions.kt
```diff
@@ -0,0 +1,156 @@
...
+
+@Serializable
+data class SynchronizationRetrieveFolderRequest(
+    val path: String,
+    val provider: String
+)
+typealias SynchronizationRetrieveFolderResponse = SynchronizedFolder
+
+
+object FileSynchronizationDescriptions: CallDescriptionContainer("files.synchronization") {
```
DanThrane, line 67:
```suggestion
object FileSynchronization : CallDescriptionContainer("files.synchronization") {
```

backend/storage-service/api/src/commonMain/kotlin/FileSynchronizationDescriptions.kt
```diff
@@ -0,0 +1,156 @@
...
+        http {
+            method = HttpMethod.Post
+
+            path {
+                using(baseContext)
+                +"folder"
+            }
+
+            body { bindEntireRequestFromBody() }
+        }
```
DanThrane, line 91:
```suggestion
        httpCreate(baseContext, "folder", roles = Roles.PRIVILEGED)
```

backend/storage-service/api/src/commonMain/kotlin/FileSynchronizationDescriptions.kt
```diff
@@ -0,0 +1,156 @@
...
+        http {
+            method = HttpMethod.Delete
+
+            path {
+                using(baseContext)
+                +"folder"
+            }
+
+            body { bindEntireRequestFromBody() }
+        }
```
DanThrane, line 110:
```suggestion
        httpDelete(baseContext, "folder", roles = Roles.PRIVILEGED)
```

This change probably requires adding the `subResource` property to `httpDelete` but it should be there since it is present on `httpCreate`.

backend/storage-service/api/src/commonMain/kotlin/FileSynchronizationDescriptions.kt
```diff
@@ -0,0 +1,156 @@
...
+        http {
+            method = HttpMethod.Post
+
+            path {
+                using(baseContext)
+                +"device"
+            }
+
+            body { bindEntireRequestFromBody() }
+        }
```
DanThrane, line 129:
```suggestion
        httpCreate(baseContext, "device", roles = Roles.PRIVILEGED)
```

backend/storage-service/api/src/commonMain/kotlin/FileSynchronizationDescriptions.kt
```diff
@@ -0,0 +1,156 @@
...
+        http {
+            method = HttpMethod.Delete
+
+            path {
+                using(baseContext)
+                +"device"
+            }
+
+            body { bindEntireRequestFromBody() }
+        }
```
DanThrane, line 148:
```suggestion
        httpDelete(baseContext, "device", roles = Roles.PRIVILEGED)
```

backend/storage-service/api/src/commonMain/kotlin/FileSynchronizationDescriptions.kt
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    val browseDevices = call<SynchronizationBrowseDevicesRequest, SynchronizationBrowseDevicesResponse,
+        CommonErrorMessage>("browseDevices") {
+        httpBrowse(baseContext, Roles.PRIVILEGED)
```
DanThrane, line 153:
```suggestion
        httpBrowse(baseContext, roles = Roles.PRIVILEGED)
```

backend/storage-service/src/jvmMain/kotlin/file/http/FileSynchronizationController.kt
```diff
@@ -0,0 +1,43 @@
...
+import dk.sdu.cloud.calls.server.securityPrincipal
+import dk.sdu.cloud.file.api.FileSynchronizationDescriptions
+import dk.sdu.cloud.file.services.SynchronizationService
+
+class SynchronizationController(
+    private val synchronizationService: SynchronizationService
+) : Controller {
+    override fun configure(rpcServer: RpcServer) = with(rpcServer) {
+        implement(FileSynchronizationDescriptions.retrieveFolder) {
+            ok(synchronizationService.retrieveFolder(Actor.SystemOnBehalfOfUser(ctx.securityPrincipal.username), request.path))
```
DanThrane, line 15:
```suggestion
            ok(synchronizationService.retrieveFolder(actorAndProject.actor, request.path))
```

This probably requires an import. I am pretty sure this functionality is present on `master`. If not, we can simply port it back.

backend/storage-service/src/jvmMain/kotlin/file/http/FileSynchronizationController.kt
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        implement(FileSynchronizationDescriptions.addFolder) {
+            synchronizationService.addFolder(Actor.SystemOnBehalfOfUser(ctx.securityPrincipal.username), request)
```
DanThrane, line 19:
```suggestion
            synchronizationService.addFolder(actorAndProject.actor, request)
```

backend/storage-service/src/jvmMain/kotlin/file/http/FileSynchronizationController.kt
```diff
@@ -0,0 +1,43 @@
...
+            ok(Unit)
+        }
+
+        implement(FileSynchronizationDescriptions.removeFolder) {
+            synchronizationService.removeFolder(Actor.SystemOnBehalfOfUser(ctx.securityPrincipal.username), request)
```
DanThrane, line 24:
```suggestion
            synchronizationService.removeFolder(actorAndProject.actor, request)
```

backend/storage-service/src/jvmMain/kotlin/file/http/FileSynchronizationController.kt
```diff
@@ -0,0 +1,43 @@
...
+            ok(Unit)
+        }
+
+        implement(FileSynchronizationDescriptions.addDevice) {
+            synchronizationService.addDevice(Actor.SystemOnBehalfOfUser(ctx.securityPrincipal.username), request)
```
DanThrane, line 29:
```suggestion
            synchronizationService.addDevice(actorAndProject.actor, request)
```

backend/storage-service/src/jvmMain/kotlin/file/http/FileSynchronizationController.kt
```diff
@@ -0,0 +1,43 @@
...
+            ok(Unit)
+        }
+
+        implement(FileSynchronizationDescriptions.removeDevice) {
+            synchronizationService.removeDevice(Actor.SystemOnBehalfOfUser(ctx.securityPrincipal.username), request)
```
DanThrane, line 34:
```suggestion
            synchronizationService.removeDevice(actorAndProject.actor, request)
```

backend/storage-service/src/jvmMain/kotlin/file/http/FileSynchronizationController.kt
```diff
@@ -0,0 +1,43 @@
...
+            ok(Unit)
+        }
+
+        implement(FileSynchronizationDescriptions.browseDevices) {
+            ok(synchronizationService.browseDevices(Actor.SystemOnBehalfOfUser(ctx.securityPrincipal.username)))
```
DanThrane, line 39:
```suggestion
            ok(synchronizationService.browseDevices(actorAndProject.actor))
```

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,205 @@
...
+                )
+            }.rows.sumOf {
+                CephFsFastDirectoryStats.getRecursiveSize(File(fsPath, it.getField(SynchronizedFoldersTable.path)))
+            }
+        }
+    }
+
+    suspend fun addFolder(actor: Actor, request: SynchronizationAddFolderRequest) {
+        // TODO Check for number of files in folder before adding (limit to 1000_000)
+        if (CephFsFastDirectoryStats.getRecursiveFileCount(File(fsPath, request.path)) > 1000_000) {
```
DanThrane, line 55:
```suggestion
        if (CephFsFastDirectoryStats.getRecursiveFileCount(File(fsPath, request.path)) > 1_000_000) {
```

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,205 @@
...
+    val user = text("user_id", notNull = true)
+}
+
+class SynchronizationService(
+    private val syncthing: SyncthingClient,
+    private val fsPath: String,
+    private val db: DBContext,
+    private val aclService: AclService
+) {
+    private suspend fun chooseFolderDevice(): LocalSyncthingDevice? {
```
DanThrane, line 34: Accept an already established session for this function. Right now this will open and close a database transaction for every single device instead of using the one that is likely already established.

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,205 @@
...
+        if (CephFsFastDirectoryStats.getRecursiveFileCount(File(fsPath, request.path)) > 1000_000) {
+            throw RPCException("Number of files in directory exceeded for synchronization", HttpStatusCode.Forbidden)
+        }
+
+        val id = UUID.randomUUID().toString()
+        val device = chooseFolderDevice()
+        val accessType = if (aclService.hasPermission(request.path, actor.username, AccessRight.WRITE)) {
+            SynchronizationType.SEND_RECEIVE
+        } else {
+            SynchronizationType.SEND_ONLY
```
DanThrane, line 64: This would allow anyone to read any file in the entire system. You need to check if they have read permissions.

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
DanThrane: This code snippet needs to be moved down below the acl check as it will otherwise leak information about files which the user does not have access to.

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
DanThrane: I suspect this code could benefit greatly from being cached for a little while.

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,205 @@
...
+                    setParameter("user", actor.username)
+                },
+                """
+                    delete from storage.synchronized_folders
+                    where id = :id and user_id = :user
+                """
+            )
+        }
+
+        syncthing.writeConfig()
```
DanThrane, line 115: Add a check to see if you really need to write configuration. For example if the code above didn't delete anything, for example in the case of lacking permissions, this will do nothing.

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    suspend fun addDevice(actor: Actor, request: SynchronizationAddDeviceRequest) {
+        if (syncthing.config.devices.find { it.id == request.id } != null) {
```
DanThrane, line 119: Can be replaced with `!syncthing.config.devices.any { it.id == request.id }`

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,205 @@
...
+                        user_id
+                    ) values (
+                        :device,
+                        :user
+                    )
+                """
+            )
+        }
+
+        syncthing.writeConfig()
```
DanThrane, line 141: Conditional write

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,205 @@
...
+                    setParameter("user", actor.username)
+                },
+                """
+                    delete from storage.user_devices
+                    where device_id = :id and user_id = :user
+                """
+            )
+        }
+
+        syncthing.writeConfig()
```
DanThrane, line 158: Conditional write

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,205 @@
...
+                """
+                        select device_id
+                        from storage.user_devices
+                        where user_id = :user
+                        limit 100
+                    """
+            ).rows.map { SynchronizationDevice(it.getField(UserDevicesTable.device)) }
+        }
+
+        return PageV2(100, devices, null)
```
DanThrane, line 176: This code doesn't actually paginate. Instead it returns 100 random user devices. It is random because you don't do any ordering of the devices.

You should look at the `db.paginateV2` call.

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,205 @@
...
+            val folder = session.sendPreparedStatement(
+                {
+                    setParameter("user", actor.username)
+                    setParameter("path", path)
+                },
+                """
+                        select id, path, device_id
+                        from storage.synchronized_folders
+                        where user_id = :user and path = :path
+                        limit 100
```
DanThrane, line 190: Update the database such that the `(user, path)` tuple is actually unique. In that case you also do not need the `limit 100` entry.

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,205 @@
...
+                    """
+            ).rows
```
DanThrane, line 192: Use `.singleOrNull()`

backend/storage-service/src/jvmMain/kotlin/file/services/SyncthingClient.kt
```diff
@@ -0,0 +1,304 @@
...
+            val result = db.withSession { session ->
+                session.sendPreparedStatement(
+                    {
+                        setParameter("localDevice", device.id)
+                    },
+                    """
+                       select id, path, access_type, d.device_id, d.user_id from storage.synchronized_folders as f
+                       inner join storage.user_devices as d on f.user_id = d.user_id
+                       where f.device_id = :localDevice
+                    """
```
DanThrane, line 256:
```suggestion
                    """
                       select id, path, access_type, d.device_id, d.user_id
                       from
                          storage.synchronized_folders f join
                          storage.user_devices d on f.user_id = d.user_id
                       where
                          f.device_id = :localDevice
                    """
```

Improved formatting and removed noise words.

backend/storage-service/src/jvmMain/kotlin/file/services/SyncthingClient.kt
```diff
@@ -0,0 +1,304 @@
...
+                                },
+                                """
+                                    select device_id
+                                    from storage.user_devices
+                                    where user_id = :user
+                                """
+                            ).rows.map { SyncthingFolderDevice(it.getField(UserDevicesTable.device)) }
+                        },
+                        path = File(fsPath, row.getField(SynchronizedFoldersTable.path)).absolutePath
+                    )
```
DanThrane, line 285: A few notes:

- Don't run queries in a loop, it is really bad for performance. Run a single query which fetches everything you need. Ideally this will be combined with your query above.
- Because you are running this across `O(n + 1)` database transactions there are no guarantees that what you found in a previous transaction hasn't already been changed.

You need to rewrite this function such that it only does one or two database queries inside a single transaction.

backend/storage-service/src/jvmMain/kotlin/file/services/SyncthingClient.kt
```diff
@@ -0,0 +1,304 @@
...
+            headers {
+                append("X-API-Key", device.apiKey)
+            }
+        }
+
+        val config = resp.content.toByteArray().toString(Charsets.UTF_8)
+        return defaultMapper.decodeFromString(config)
+    }
+
+    suspend fun writeConfig() {
```
DanThrane, line 245: This function is called every time any change is made. Changes are made as a result of RPCs caused by a user action. This means that you have _many_ different threads hitting this call, yet there is currently no locking to prevent multiple instances of this function going off at the same time. You need to add some kind of locking to this function (e.g. `Mutex`).

backend/storage-service/src/jvmMain/resources/db/migration/V13__Synchronization.sql
```diff
@@ -0,0 +1,15 @@
+create table synchronized_folders (
+    id          varchar(36) not null primary key,
+    device_id   varchar(64) not null,
+    path        text        not null,
+    access_type varchar(20) not null default 'SEND_RECEIVE',
```
DanThrane, line 5: Seems like a dangerous default to me.

DanThrane (changes requested):

DanThrane (changes requested):
A few minor adjustments. Will check later once we have finished our other discussion.

backend/storage-service/src/jvmMain/kotlin/file/services/SyncthingClient.kt
```diff
@@ -0,0 +1,312 @@
...
+                           select id, path, access_type, f.device_id as local_device_id, d.device_id, d.user_id, f.user_id
+                           from
+                              storage.synchronized_folders f join
+                              storage.user_devices d on f.user_id = d.user_id
+                        """
+                )
+            }.rows
+
+            config.devices.forEach { device ->
+                val oldConfig = readConfig(device)
```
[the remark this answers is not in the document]
xiroV: We have to write each syncthing-config individually (if you are referring to the loop) so I don't see another way to do it. Maybe we can skip the `readConfig` call, I just have to make sure that the default values are correct.

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,257 @@
...
+
+                val accessType = if (aclService.hasPermission(folder.path, actor.username, AccessRight.WRITE)) {
+                    SynchronizationType.SEND_RECEIVE
+                } else if (aclService.hasPermission(folder.path, actor.username, AccessRight.READ)) {
+                    SynchronizationType.SEND_ONLY
+                } else {
+                    throw RPCException.fromStatusCode(HttpStatusCode.Unauthorized)
+                }
+
+                session.sendPreparedStatement(
```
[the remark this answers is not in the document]
xiroV: I agree, but I'm not completely sure about the best way to do it. It doesn't seem like we do single-query bulk inserts anywhere else (on master).

Also, it isn't possible to add folders in bulks from the frontend, so currently it isn't an issue.

xiroV:
Converted to draft while issue #2568 (and surrounding issues) are being resolved.

backend/sync-mounter-service/src/jvmMain/kotlin/Main.kt
```diff
@@ -0,0 +1,26 @@
+package dk.sdu.cloud.sync.mounter
+
+import dk.sdu.cloud.auth.api.AuthenticatorFeature
+import dk.sdu.cloud.micro.*
+import dk.sdu.cloud.service.CommonServer
+import dk.sdu.cloud.sync.mounter.api.SyncMounterServiceDescription
+
+data class SyncMounterConfiguration(
+    val cephfsBaseMount: String? = null,
+    val syncBaseMount: String? = null,
```
DanThrane, line 10: Use defaults which match the hard-coded values you have all over the code. e.g. `/mnt/sync` should be the default value of `syncBaseMount`.

backend/sync-mounter-service/src/jvmMain/kotlin/Main.kt
```diff
@@ -0,0 +1,26 @@
...
+    val deviceId: String? = null
```
DanThrane, line 11: Is a value of `null` okay?

backend/sync-mounter-service/src/jvmMain/kotlin/Server.kt
```diff
@@ -0,0 +1,93 @@
...
+                MountController(mountService)
+            )
+        }
+        
+        startServices()
+    }
+
+    override fun onKtorReady() {
+        runBlocking {
+            val readyFile = File(joinPath(config.syncBaseMount ?: "/mnt/sync", "ready"))
```
DanThrane, line 42: See comment about config.

backend/sync-mounter-service/src/jvmMain/kotlin/Server.kt
```diff
@@ -0,0 +1,93 @@
...
+                        MountFolder(folder.id, folder.path)
+                    }
+                )
+            )
+
+            readyFile.createNewFile()
+        }
+    }
+
+    override fun stop() {
```
DanThrane, line 71: Replace with a shutdown hook instead, `stop()` isn't called quite as reliably as the shutdown hook would be.

backend/sync-mounter-service/src/jvmMain/kotlin/Server.kt
```diff
@@ -0,0 +1,93 @@
...
+        runBlocking {
+            val mountService = MountService(config)
+
+            File(joinPath(config.syncBaseMount ?: "/mnt/sync", "ready")).delete()
+
+            val syncDir = File(joinPath(config.syncBaseMount ?: "/mnt/sync"))
+
+            mountService.mount(
```
DanThrane, line 79: Shouldn't this be an unmount?

backend/sync-mounter-service/src/jvmMain/kotlin/Server.kt
```diff
@@ -0,0 +1,93 @@
...
+                MountRequest(
+                    syncDir.listFiles().map { file ->
+                        MountFolder(file.name, file.absolutePath)
+                    }
+                )
+            )
+
+            syncDir.listFiles().forEach { file ->
```
DanThrane, line 87: This is probably not needed and a bit risky.

backend/sync-mounter-service/src/jvmMain/kotlin/services/MountService.kt
```diff
@@ -0,0 +1,102 @@
...
+                    fileDescriptors[i] = CLibrary.INSTANCE.openat(fileDescriptors[i - 1], components[i], O_NOFOLLOW, 0)
+
+                    if (fileDescriptors[i] < 0) {
+                        throw RPCException.fromStatusCode(HttpStatusCode.InternalServerError)
+                    }
+
+                    CLibrary.INSTANCE.close(previousFd)
+                }
+            } catch (ex: Throwable) {
+                fileDescriptors.closeAll()
```
DanThrane, line 52: Should be done in a `finally` block

backend/sync-mounter-service/src/jvmMain/kotlin/services/MountService.kt
```diff
@@ -0,0 +1,102 @@
...
+                throw ex
+            }
+            CLibrary.INSTANCE.close(fileDescriptors.last())
+
+            CLibrary.INSTANCE.mount(source.absolutePath, target.absolutePath, null, MS_BIND, null)
```
DanThrane, line 57: Race condition is still present because of this call. You need to bind mount the file descriptor and not the original path. It should be present in `/proc/self/fd/$FD`.

backend/sync-mounter-service/src/jvmMain/kotlin/services/MountService.kt
```diff
@@ -0,0 +1,102 @@
...
+            if (!source.exists() || !source.isDirectory) {
+                throw RPCException.fromStatusCode(HttpStatusCode.NotFound, "Invalid source")
+            }
+
+            val target = File(joinPath(config.syncBaseMount ?: "/mnt/sync", item.id))
+            if (target.exists()) {
+                unmount(UnmountRequest(listOf(MountFolderId(item.id))))
+            }
+
+            if (!target.mkdir()) {
```
DanThrane, line 27: I think we need some input validation on the source and target. Make sure they are actually children of the folder you expect.

backend/sync-mounter-service/src/jvmMain/kotlin/services/MountService.kt
```diff
@@ -0,0 +1,102 @@
...
+            CLibrary.INSTANCE.close(fileDescriptors.last())
+
+            CLibrary.INSTANCE.mount(source.absolutePath, target.absolutePath, null, MS_BIND, null)
+        }
+
+        return MountResponse
+    }
+
+    fun unmount(request: UnmountRequest): UnmountResponse {
+        request.items.forEach { item ->
```
DanThrane, line 64: Missing input validation, similar to `mount`.

backend/sync-mounter-service/src/jvmMain/kotlin/services/MountService.kt
```diff
@@ -0,0 +1,102 @@
...
+            if (!target.delete()) {
+                throw RPCException.fromStatusCode(HttpStatusCode.InternalServerError, "Failed to delete target")
+            }
+        }
+
+        return UnmountResponse
+    }
+
+    fun ready(): ReadyResponse {
+        if (File(joinPath(config.syncBaseMount ?: "/mnt/sync", "ready")).exists()) {
```
DanThrane, line 81: You probably don't need to query the file here. The file is mostly needed to delay starting syncthing. From the service you could use an atomic variable, this would be significantly faster than reading from the filesystem.

DanThrane (changes requested):
Added review for the syncthing-mounter, still missing review of rest. Will review after fix of comments.

backend/sync-mounter-service/src/jvmMain/kotlin/services/MountService.kt
xiroV: We can check if `source` starts with `cephfsBase + "home"`, is this what you mean?
And in that case, `target` is already made from `syncBase + itemId`, what should we validate?

DanThrane: You need to check that the _resolved_ source is inside the expected folder. The biggest problem is if the source service starts acting malicious and tries to mount arbitrary folders by using relative paths. For example, "$cephfsBase/home/foo/bar/../../../../../../something/else/entirely` still starts with the expected path but will resolve to something completely different.

backend/sync-mounter-service/src/jvmMain/kotlin/Server.kt
xiroV: I've moved it to a shutdownHook in `start`, is this the correct way to do it?

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,308 @@
...
+                if (!internalFile.exists() || !internalFile.isDirectory) {
+                    throw RPCException.fromStatusCode(HttpStatusCode.NotFound)
+                }
+
+                val syncType = if (aclService.hasPermission(folder.path, actor.username, AccessRight.WRITE)) {
+                    SynchronizationType.SEND_RECEIVE
+                } else if (aclService.hasPermission(folder.path, actor.username, AccessRight.READ)) {
+                    SynchronizationType.SEND_ONLY
+                } else {
+                    throw RPCException.fromStatusCode(HttpStatusCode.Unauthorized)
```
DanThrane, line 83: Forbidden would technically be the correct error code to use here.

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,308 @@
...
+            """
+            ).rowsAffected
+        }
+
+        affectedDevices.forEach { device ->
+            val mounter = Mounts.ready.call(Unit, authenticatedClient)
+            if (mounter.statusCode != HttpStatusCode.OK || !mounter.orThrow().ready || !syncthing.isReady(device)) {
+                throw RPCException(
+                    "The synchronization feature is offline at the moment. Your folders will be synchronized when it returns online.",
+                    HttpStatusCode.InternalServerError
```
DanThrane, line 142: 503 Service Unavailable or 502 Bad gateway seem more appropriate.

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,308 @@
...
+                    setParameter("user", actor.username)
+                },
+                """
+                    delete from storage.synchronized_folders
+                    where id in (select unnest(:ids::text[])) and user_id = :user
+                """
+            ).rowsAffected
+        }
+
+        Mounts.unmount.call(
```
DanThrane, line 166: Exception is thrown outside of the transaction, this will probably give some very confusing results. Also what happens if the mounter is down in this case?

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,308 @@
...
+                        unnest(:users::text[]),
+                        unnest(:access::text[])
+                    )
+            """
+            ).rowsAffected
+        }
+
+        affectedDevices.forEach { device ->
+            val mounter = Mounts.ready.call(Unit, authenticatedClient)
+            if (mounter.statusCode != HttpStatusCode.OK || !mounter.orThrow().ready || !syncthing.isReady(device)) {
```
DanThrane, line 139: I would probably add a comment here to explain the difference between the mounter being ready and syncthing being ready.

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,308 @@
...
+        if (!Roles.PRIVILEGED.contains(principal.role)) {
+            throw RPCException.fromStatusCode(HttpStatusCode.Unauthorized)
+        }
+
+        return db.withSession { session ->
+            session.sendPreparedStatement(
+                {
+                    setParameter("device", request.device)
+                },
+                """
```
DanThrane, line 193: Indentation

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
```diff
@@ -0,0 +1,308 @@
...
+            throw RPCException.fromStatusCode(HttpStatusCode.InternalServerError)
+        }
+
+        if (affectedRows > 0) {
+            syncthing.writeConfig()
+        }
+    }
+
+    suspend fun browseFolders(
+        principal: SecurityPrincipal,
```
DanThrane, line 181: Prefer using `Actor`

backend/storage-service/src/jvmMain/kotlin/file/services/acl/AclService.kt
```diff
@@ -287,6 +296,110 @@ class AclService(
...
+        )
+
+        db.withSession { session ->
+            val syncedChildren = session.sendPreparedStatement(
+                {
+                    setParameter("path", path)
+                },
+                """
+                    select id, user_id, path, access_type from storage.synchronized_folders
+                    where path like :path || '%'
```
DanThrane, line 315: This will cause problems. This will end up catching `/home/foo/bar` and `/home/foobar/baz` if the path is `/home/foo`.

backend/storage-service/src/jvmMain/kotlin/file/services/acl/AclService.kt
```diff
@@ -287,6 +296,110 @@ class AclService(
...
+                            select
+                                unnest(:ids::text[]) as id,
+                                unnest(:users::text[]) as user,
+                                unnest(:access::text[]) as access
+                        ) as new
+                        where f.id = new.id and f.user_id = new.user
+                    """
+                )
+
+                Mounts.mount.call(
```
DanThrane, line 370: Does the `mount` call correctly handling this potentially already being present?

backend/storage-service/src/jvmMain/kotlin/file/services/acl/AclService.kt
DanThrane: You also need to make sure that you are only storing the _normalized_ version of paths in the `synchronize_folders` table. You might already be doing this but you should check.

backend/storage-service/src/jvmMain/kotlin/file/services/acl/AclService.kt
```diff
@@ -127,6 +134,8 @@ class AclService(
                 ) as JsonObject
             )
         )
+
+        updateSynchronizationType(request.path)
```
DanThrane, line 138: Path definitely needs to be normalized here.

DanThrane (changes requested):
Some more changes

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
xiroV: I've updated it to do the write to Syncthing before attempting to unmount. I think if the mounter is down, it's important that Syncthing at least is in the right state.

backend/storage-service/src/jvmMain/kotlin/file/services/acl/AclService.kt
xiroV: Mount will attempt to unmount any folders which already exist, before attempting to mount.

However, the call to mount in AclService shouldn't actually be there at all, since the list of folders to mount will always be empty. I have removed it.

Reason: If permissions change such that a user no longer have (at least) read permission, the folder will be removed completely from syncthing and unmounted. If the user regains permission to the directory, we cannot remount it since we have removed the association completely.

We can change this such that it doesn't delete the entry from the database (like it does now), but just from Syncthing + unmount. In that case we can remount the folder if the user regains permission to the folder.

backend/sync-mounter-service/src/jvmMain/kotlin/services/MountService.kt
```diff
@@ -0,0 +1,128 @@
...
+
+class MountService(
+    val config: SyncMounterConfiguration,
+    val ready: AtomicBoolean
+) {
+    private val MS_BIND: Long = 4096
+    private val O_NOFOLLOW = 0x20000
+
+    @OptIn(ExperimentalPathApi::class)
+    fun mount(request: MountRequest) {
```
DanThrane, line 22: Same issue with `startsWith`. You are not checking if they are a descendant but just checking if the strings starts with another. You end up with this issue:

```
>>> java.io.File("/mnt/cephfssomethingelse").canonicalPath.startsWith("/mnt/cephfs")
res2: kotlin.Boolean = true
```

backend/sync-mounter-service/src/jvmMain/kotlin/services/MountService.kt
```diff
@@ -0,0 +1,128 @@
...
+                    fileDescriptors[i] = CLibrary.INSTANCE.openat(fileDescriptors[i - 1], components[i], O_NOFOLLOW, 0)
+
+                    if (fileDescriptors[i] < 0) {
+                        throw RPCException.fromStatusCode(HttpStatusCode.InternalServerError)
+                    }
+
+                    CLibrary.INSTANCE.close(previousFd)
+                }
+
+                val realSourcePath = Paths.get("/proc/self/fd/${fileDescriptors.last()}").readSymbolicLink()
```
DanThrane, line 69: Don't read the symbolic link as this will immediately reintroduce the bug.

DanThrane (changes requested):

backend/storage-service/src/jvmMain/kotlin/file/services/SynchronizationService.kt
xiroV: I've changed this again, by just moving the unmount call inside the transaction.

backend/sync-mounter-service/src/jvmMain/kotlin/services/MountService.kt
xiroV: The joinPath already adds a ending slash to the path. I've updated the paths from the configuration to always end with a slash.

DanThrane (approved):
Approved for internal testing. Will need additional PRs for a full release.

PR #2598, opened 2021-08-06
Fixes #2533

### LINKED ISSUE #2533: [storage] Select different drive from the "file selector" UI

Tiggles:


DanThrane (changes requested):
I think this is a good start but we probably need to expand it with some kind of pagination. Otherwise this code will only work if the user has less than 100 drives.

DanThrane (approved):

PR #2615, opened 2021-08-13
Fixes #2608

### LINKED ISSUE #2608: [storage] Embedded FileTables can navigate to filepreview

Tiggles:
Fixes #2608


DanThrane (commented):
Isn't this missing a case for when we are embedded? I believe the properties page (which include the preview) already works if you are embedded.

Tiggles:
It's intended to not allow navigation for files when embedded.

What is the use-case that should work that you are referring to?

DanThrane (approved):

PR #2688, opened 2021-09-01
Fixes #2687

### LINKED ISSUE #2687: [storage] File favorite deletion sometimes uses an empty string in the request
I'm guessing that this is related to improper use of state or potentially a lack of refresh.

Tiggles:
Reverts some template names in the FE code to match auto-generated template names from migration scripts.

### DIFF
--- a/frontend-web/webclient/app/Dashboard/Dashboard.tsx
+++ b/frontend-web/webclient/app/Dashboard/Dashboard.tsx
@@ -104,7 +104,7 @@ function Dashboard(props: DashboardProps & {history: History}): JSX.Element {
         }));
         fetchFavoriteFiles(metadataApi.browse({
             filterActive: true,
-            filterTemplate: "Favorite",
+            filterTemplate: "UCloud favorite",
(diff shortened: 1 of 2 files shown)


frontend-web/webclient/app/Dashboard/Dashboard.tsx
```diff
@@ -125,7 +125,7 @@ function Dashboard(props: DashboardProps & {history: History}): JSX.Element {
                 favoriteFiles={favoriteFiles}
                 onDeFavorite={() => fetchFavoriteFiles(metadataApi.browse({
                     filterActive: true,
-                    filterTemplate: "Favorite",
+                    filterTemplate: "UCloud favorite",
```
Tiggles, line 128: A specific reference to UCloud in the templates could be an issue, correct?

DanThrane:
You are probably right. I also didn't intend to make it use the `isFavorite` property. I had actually changed this manually in my own database but forgot to apply in the migration script. I'll fix the script and close this PR since we won't need it.

PR #2716, opened 2021-09-15
Storage vite

Tiggles:


frontend-web/webclient/vite.config.ts
```diff
@@ -0,0 +1,70 @@
...
+        case "development":
+            return `https://${DEV_SITE}`;
+        case "compose":
+            return "http://backend:8080";
+        case "local-dev":
+        case "production":
+        default:
+            return "localhost:8080";
+    }
+}
```
Tiggles, line 21: I want to make sure these targets are correct.

`development` is `npm run start`
`compose` is `npm run start:compose`
`local-dev` is `npm run start:local`
`production` is `npm run build`

frontend-web/webclient/vite.config.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export default ({mode, ...rest}: {mode: Mode; command: string}): UserConfigExport => {
+    const sharedProxySetting = {
+        target: targetFromConfig(mode),
+        ws: true,
+        changeOrigin: true,
+        headers: {
+            "Access-Control-Allow-Origin": "*",
+            "Access-Control-Allow-Methods": "GET, POST, PUT, DELETE, PATCH, OPTIONS",
+            "Access-Control-Allow-Headers": "X-Requested-With, content-type, Authorization"
```
Tiggles, line 31: I think these are correct. Please validate.

Tiggles:
First of all: sorry.

What this PR does:

- '@/' prefix is added to reference our own files, when we don't want to use relative paths.
- Vite is added
- Webpack is removed
- React Markdown is updated (minor thing, but it crashed the whole site)
- `run` commands are changed. Suffixed by target platform, e.g. `run:compose` or `run:local`
- For some reason, VSC assumes the context is `node`, so `setTimeout` has been changed to `window.setTimeout`. This also fixed a bug for one of these functions.
- Font has been changed to `inter`. This is something @claudiopica brought up a while ago. I'm not sure this is the best file format, as it is about 700 KB.

DanThrane (approved):

Issue #2927, opened 2021-11-18
[2021.3.0] Search-button is clickable when browser window is narrow


xiroV:
When resizing the browser window, the search-bar will at some point disappear, and a search button will appear instead.

The button is clickable even on pages where the search field is disabled, and leads to the "old" Files/Applications search page.

on dev.

Tiggles:
How this works,.

PR #2933, opened 2021-11-19
Hide search icon on small displays

xiroV:
Wasn't working, however we might want to come up with a solution such that users can still search on small displays.

When clicking the search icon it just led to a "Page not found" (probably the "old" file search page).


Tiggles:
I'm working on a solution for this.

xiroV:
fixes #2927

Tiggles:
That adds searching for smaller screens, that is.

PR #3050, opened 2021-12-01
Storage scan

### LINKED ISSUE #2824: [2021.3.0] File search not implemented
File search does not seem to work. Returns 404 in all cases.

Also when searching from within a drive, it deselects the current drive. This might be intended behavior though, if search is intended to work across drives.

hschu12:
Scan and search of file system implemented 
Fixes #2824


backend/file-orchestrator-service/api/src/commonMain/kotlin/PathUtils.kt
```diff
@@ -31,6 +31,8 @@ fun String.components(): List<String> = removePrefix("/").removeSuffix("/").spli
 
 fun String.fileName(): String = substringAfterLast('/')
 
+fun String.depth(): Int = split("/").size - 1
```
Tiggles, line 34: I don't know if it will ever be the case, but won't this be wrong for paths like `/foo/`?

Something like
```
fun String.depth(): Int = split("/").filter {it.size > 0}.size - 1
```
Would fix this, but I'm not sure that would ever be the case.

Tiggles: Actually, no. My case won't work either with that code snippet.

backend/file-ucloud-service/api/src/commonMain/kotlin/QueryDescriptions.kt
```diff
@@ -0,0 +1,274 @@
+package dk.sdu.cloud.file.ucloud.api
```
DanThrane, line 1: Is this even in use?

backend/file-ucloud-service/src/jvmMain/kotlin/services/ElasticQueryService.kt
```diff
@@ -0,0 +1,490 @@
...
+import org.elasticsearch.index.query.QueryBuilder
+import org.elasticsearch.index.query.QueryBuilders
+import org.elasticsearch.index.query.RangeQueryBuilder
+import org.elasticsearch.index.query.TermsQueryBuilder
+import org.elasticsearch.search.aggregations.AggregationBuilders
+import org.elasticsearch.search.aggregations.Aggregations
+import org.elasticsearch.search.aggregations.metrics.*
+import org.elasticsearch.search.builder.SearchSourceBuilder
+import org.elasticsearch.search.sort.SortOrder
+import services.FileScanner
```
DanThrane, line 36: Wrong package

backend/file-ucloud-service/src/jvmMain/kotlin/services/FileScanner.kt
```diff
@@ -0,0 +1,505 @@
+package services
```
DanThrane, line 1: Package is wrong

backend/file-ucloud-service/src/jvmMain/kotlin/services/FileScanner.kt
```diff
@@ -0,0 +1,505 @@
...
+                        withContext(pool) {
+                            launch {
+                                val file =
+                                    pathConverter.relativeToInternal(RelativeInternalFile("/projects/$path}"))
+                                submitScan(file, coll)
+                            }.join()
+                        }
+                    }
+
+                    if (resolvedCollections.next == null) {
```
DanThrane, line 216: Won't this always be the case if we just do `itemsPerPage = 50`?

backend/file-ucloud-service/src/jvmMain/kotlin/services/FileScanner.kt
```diff
@@ -0,0 +1,505 @@
...
+                elastic.get(GetRequest(FILES_INDEX, pathConverter.internalToUCloud(file).path), RequestOptions.DEFAULT)?.sourceAsString
+            if (source != null) {
+                defaultMapper.decodeFromString<ElasticIndexedFile>(source)
+            } else {
+                null
+            }
+        }
+
+        var newUpperLimitOfEntries = 1L
+        if (File(file.path).isDirectory) {
```
DanThrane, line 273: Should be changed to use the `NativeFs` interface instead (to correctly handle symlinks)

backend/file-ucloud-service/src/jvmMain/kotlin/services/FileScanner.kt
```diff
@@ -0,0 +1,505 @@
...
+                    null
+                }
+            }
+        }.joinAll()
+    }
+
+    data class ShouldContinue(val shouldContinue: Boolean, val newUpperLimitOfEntries: Long)
+
+    private fun shouldContinue(internalFile: InternalFile, upperLimitOfEntries: Long): ShouldContinue {
+        val file = File(internalFile.path)
```
DanThrane, line 371: Should be changed to use the `NativeFS` interface instead

DanThrane (changes requested):
Mostly looks good to me. I have added a few comments, we at the very least need to remove all traces of the Java File API, since it doesn't handle the symlinks correctly.

backend/file-ucloud-service/api/src/commonMain/kotlin/QueryDescriptions.kt
hschu12: QueryDescriptions endpoint is not (and is now deleted) but the FileQuery etc. is used in scan.

backend/file-ucloud-service/src/jvmMain/kotlin/services/FileScanner.kt
hschu12: I also expected this to alway be null, but this is the place where sometimes it runs one more turn.
I can delete it if you want

backend/file-orchestrator-service/api/src/commonMain/kotlin/PathUtils.kt
hschu12: based on a normalized path this is correct . Have however made an if on (equals("/")) since this should be depth 0 according to doc

backend/file-ucloud-service/src/jvmMain/kotlin/services/FileScanner.kt
DanThrane: We can keep it, but I suspect it is not needed.

DanThrane (approved):

Issue #3168, opened 2022-01-06
Bottom Project Resource dropdown goes out of viewport


Tiggles:
![Screenshot 2022-01-06 at 09 46 21](https://user-images.githubusercontent.com/7610257/148356151-6b3ff2a6-3d16-4757-973e-b31060a6133a.png)

PR #3180, opened 2022-01-06
Temp fix for #3168

Tiggles:


Tiggles:
Expand bottom box, so it doesn't go out of viewport. Should instead move Clickable dropdown above cursor in the longterm.

DanThrane (approved):

PR #3192, opened 2022-01-13
Issue 3185

Tiggles:
New:
image
Old: 
image


DanThrane (changes requested):
Just did a quick look and the UI looks good to me. But this should replace the old description page entirely and I don't see any code which removes the old page. Just to be clear I am talking about a page such as; https://cloud.sdu.dk/app/applications/details/superset/1.3.0/

Tiggles:
I was going to ask if that was going to be removed, as the new page seems to have everything the old one has. The only thing it's missing is a better version view:
image

Tiggles:
I think that's everything done that we talked about, except a different way of showing you can click recent jobs and I have not merged Machine Type and node count, as these are of different height.

DanThrane (approved):

PR #3286, opened 2022-02-09
Datacollection

hschu12:
New service. Mostly just copied from previous master. To next report many Queries need to be changed, but there might be changes to format so it will be done next time


DanThrane (approved):
A few things I would like to change but I am also aware that this is probably the last time we run it like this.

PR #3306, opened 2022-02-17
Syncthing fixes

### LINKED ISSUE #3278: [sync] Requests to Syncthing are slow
Currently on dev requests where a write to Syncthing is required, the request takes at least 6 seconds.

xiroV:
### Implementation of new  Syncthing calls (fixes #3278)
Currently we are using a single main-call for (almost) all communication with Syncthing (`writeConfig`) which will fetch all devices and folders, generate a Syncthing config which will then be sent to Syncthing. However, requesting Syncthing to overwrite its config appears to make it restart (including the API server). Because of this we experience a lot of communication problems between the kTor client and Syncthing.

This PR includes implementations for less drastic calls to Syncthing, meaning that we will instead only use `writeConfig` when the service (provider) is started. Instead it will be left to Syncthing to generate/update its configuration internally using the following calls: `addFolders`, `addDevices`, `removeFolders`, `removeDevices`.

`addFolders` and `addDevices` will still fetch all folders and devices respectively, and send these to Syncthing. This is to minimize the risk of the provider and Syncthing going too much out of sync.

`removeFolders` and `removeDevices` will **only remove the requested folders and devices** respectively. An improvement to this would be to first fetch the list of folders and devices from the provider, then from Syncthing, and finally ask Syncthing to remove the difference. This is not currently implemented.

### Propagate errors properly back to orchestrator
Fixes the issue with orchestrator and provider going out of sync by properly propagating errors back to the orchestrator.

### Only rescan syncthing on startup
Currently we ask Syncthing to rescan when the service is started. We keep repeating the request (in intervals) until we get a response. However, Syncthing will only respond once it's done rescanning (which can take long). Thus, we currently keep asking Syncthing to rescan.

This PR removes any retrying of the rescan request, thus we only ask Syncthing to rescan once when the service is started, and we ignore any bad responses. In the best case Syncthing will get the request, and rescan, and otherwise it will do the rescan automatically at a later point (when the automatic rescan sets in).

Syncthing by itself does not block other requests while rescanning, but currently no requests seem to reach Syncthing. Thus, there's a chance we are blocking requests to Syncthing from the provider due to locks being occupied by the rescanning. 

This bug have only just appeared because of the high number of files added to sync.

### DIFF
--- a/backend/file-ucloud-service/src/jvmMain/kotlin/services/SyncthingClient.kt
+++ b/backend/file-ucloud-service/src/jvmMain/kotlin/services/SyncthingClient.kt
@@ -21,7 +21,6 @@ import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.encodeToString
-import java.io.File
 import java.net.ConnectException
 import java.util.concurrent.atomic.AtomicLong
 
@@ -257,7 +256,7 @@ class SyncthingClient(
             log.debug("Writing config to Syncthing")
 
             // NOTE(Brian): We are waiting for changes to be written to all syncthing devices, sending requests every
-            // 5 seconds, for 30 seconds. Syncthing will start rejecting all requests if we send too many in a row.
+            // 5 seconds, for 30 seconds.
             while (pendingDevices.isNotEmpty()) {
                 if (Time.now() > lastWrite.get() + 5_000) {
                     mutex.withLock {
@@ -311,10 +310,7 @@ class SyncthingClient(
                                                 }.map {
                                                     SyncthingFolderDevice(it.getField(SyncDevicesTable.device))
                                                 },
-                                            path = File(
-                                                "/mnt/sync",
-                                                row.getField(SyncFoldersTable.id).toString()
-                                            ).absolutePath,
+                                            path = "/mnt/sync/${row.getField(SyncFoldersTable.id)}",
                                             type = SynchronizationType.valueOf(row.getField(SyncFoldersTable.syncType)).syncthingValue,
                                             rescanIntervalS = device.rescanIntervalSeconds
                                         )
@@ -341,7 +337,7 @@ class SyncthingClient(
                                     }
                                 }
 
-                                if (resp.status != HttpStatusCode.OK) {
+                                if (!resp.status.isSuccess()) {
                                     throw RPCException(
                                         resp.content.toByteArray().toString(Charsets.UTF_8),
                                         dk.sdu.cloud.calls.HttpStatusCode.BadRequest
@@ -400,7 +396,7 @@ class SyncthingClient(
                             }
                         }
 
-                        if (resp.status != HttpStatusCode.OK) {
+                        if (!resp.status.isSuccess()) {
                             throw RPCException(
                                 resp.content.toByteArray().toString(Charsets.UTF_8),
                                 dk.sdu.cloud.calls.HttpStatusCode.BadRequest
@@ -443,9 +439,9 @@ class SyncthingClient(
     suspend fun rescan(devices: List<LocalSyncthingDevice> = emptyList()) {
         if (lock.acquire()) {
             mutex.withLock {
-                val pendingDevices = devices.ifEmpty { config.devices }
+                val localDevices = devices.ifEmpty { config.devices }
                 log.info("Attempting rescan of syncthing")
-                pendingDevices.forEach { device ->
+                localDevices.forEach { device ->
                     try {
                         httpClient.post<HttpResponse>(deviceEndpoint(device, "/rest/db/scan")) {
                             headers {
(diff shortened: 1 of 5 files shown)


xiroV:
Latest commit should also fix the issue with orchestrator and provider going out of sync by properly propagating errors back to the orchestrator.

backend/file-ucloud-service/src/jvmMain/kotlin/services/SyncthingClient.kt
```diff
@@ -443,28 +443,316 @@ class SyncthingClient(
...
+                                {
+                                    setParameter("devices", pendingDevices.map { it.id })
+                                },
+                                """
+                                       select f.id, f.path, f.sync_type, f.device_id as local_device_id, d.device_id, d.user_id, f.user_id
+                                       from
+                                          file_ucloud.sync_folders f join
+                                          file_ucloud.sync_devices d on f.user_id = d.user_id
+                                       where
+                                          f.device_id in (select unnest(:devices::text[]))
```
DanThrane, line 488: Could be improved with `f.device_id = some(:devices::text[])`

backend/file-ucloud-service/src/jvmMain/kotlin/services/SyncthingClient.kt
```diff
@@ -443,28 +443,316 @@ class SyncthingClient(
...
+                                                    it.getField(SyncFoldersTable.id) == row.getField(
+                                                    SyncFoldersTable.id
+                                                ) &&
+                                                    it.getField(SyncDevicesTable.user) == row.getField(
+                                                    SyncFoldersTable.user
+                                                )
+                                            }.map {
+                                                SyncthingFolderDevice(it.getField(SyncDevicesTable.device))
+                                            },
+                                        path = File(
```
DanThrane, line 515: I would avoid using `File` for this.

backend/file-ucloud-service/src/jvmMain/kotlin/services/SyncthingClient.kt
```diff
@@ -443,28 +443,316 @@ class SyncthingClient(
...
+                                        defaultMapper.encodeToString(newFolders),
+                                        ContentType.Application.Json
+                                    )
                                     headers {
                                         append("X-API-Key", device.apiKey)
                                     }
                                 }
 
-                                if (resp.status == HttpStatusCode.OK) {
+                                if (resp.status != HttpStatusCode.OK) {
```
DanThrane, line 535: Technically speaking this could reject requests which were successful but didn't return explicitly 200 (e.g. anything between 201-299). Seems unlikely though.

backend/file-ucloud-service/src/jvmMain/kotlin/services/SyncthingClient.kt
```diff
@@ -443,28 +443,316 @@ class SyncthingClient(
...
+                            } catch (ex: RPCException) {
+                                throw RPCException("Invalid Syncthing Configuration", dk.sdu.cloud.calls.HttpStatusCode.BadRequest)
+                            } catch (ex: Throwable) {
+                                // Do nothing
+                            }
+                        }
+                    }
+                    lastWrite.set(Time.now())
+
+                    if (Time.now() > timeout) {
```
DanThrane, line 553: I am not entirely sure what the timeout does. As far as I can tell, we wait for the request to finish regardless. So throwing at this point, when it has finally completed, seems wasteful.

backend/file-ucloud-service/src/jvmMain/kotlin/services/SyncthingClient.kt
```diff
@@ -443,28 +443,316 @@ class SyncthingClient(
...
+                        throw RPCException(
+                            "Unable to contact one or more syncthing clients",
+                            dk.sdu.cloud.calls.HttpStatusCode.ServiceUnavailable
+                        )
+                    }
+                }
+            }
+            lock.release()
```
DanThrane, line 561: Lock is never released if an exception is thrown above.

backend/file-ucloud-service/src/jvmMain/kotlin/services/SyncthingClient.kt
```diff
@@ -443,28 +443,316 @@ class SyncthingClient(
...
+                mutex.withLock {
+                    deviceFolders.value.forEach { folder ->
+                        try {
+                            val resp = httpClient.delete<HttpResponse>(deviceEndpoint(deviceFolders.key, "/rest/config/folders/$folder")) {
+                                headers {
+                                    append("X-API-Key", deviceFolders.key.apiKey)
+                                }
+                            }
+
+                            if (resp.status != HttpStatusCode.OK) {
```
DanThrane, line 579: `resp.status.isSuccess()`

backend/file-ucloud-service/src/jvmMain/kotlin/services/SyncthingClient.kt
```diff
@@ -443,28 +443,316 @@ class SyncthingClient(
...
-                            // do nothing
+                            log.error("Syncthing responded: ${ex.message}")
+                            throw RPCException(
+                                "Invalid Syncthing Configuration",
+                                dk.sdu.cloud.calls.HttpStatusCode.BadRequest
+                            )
+                        }
+                    }
+                }
+                lock.release()
```
DanThrane, line 594: Lock is never released if an exception is thrown.

backend/file-ucloud-service/src/jvmMain/kotlin/services/SyncthingClient.kt
```diff
@@ -443,28 +443,316 @@ class SyncthingClient(
...
+
+                    if (Time.now() > timeout) {
+                        throw RPCException(
+                            "Unable to contact one or more syncthing clients",
+                            dk.sdu.cloud.calls.HttpStatusCode.ServiceUnavailable
+                        )
+                    }
+                }
+            }
+            lock.release()
```
DanThrane, line 677: Lock is not guaranteed to be released.

DanThrane (changes requested):
I think there are a few things that should be changed. I would also like to have a few comments in the source code about how the timeouts are supposed to work, since we seem to always be waiting for a response from syncthing, regardless of what the timeout value is equal to.

DanThrane (approved):

PR #3429, opened 2022-05-05
Fix Syncthing interface becoming inaccessible after restart

### LINKED ISSUE #3414: Syncthing interface is inaccessible

xiroV:
fixes #3414 

Issue seemed to be caused by `ApplicationProxyService` always using an existing tunnel for envoy. Thus, upon restart a new IP would be assigned to the application, but the Envoy configuration would never be updated.

This PR changes how the `ApplicationProxyService` and `TunnelManager` works, by always re-creating the tunnel. I'm unsure if this could cause any other problems though, but it seems to fix the issue.


backend/app-kubernetes-service/src/jvmMain/kotlin/services/proxy/TunnelManager.kt
```diff
@@ -55,36 +55,28 @@ class TunnelManager(private val k8: K8Dependencies) {
         cleanupExecutor.shutdownNow()
     }
 
-    suspend fun createOrUseExistingTunnel(id: String, remotePort: Int, rank: Int): Tunnel {
+    suspend fun createOrRecreateTunnel(id: String, remotePort: Int, rank: Int): Tunnel {
```
DanThrane, line 58: Ideally this should only be recreated if the IP has changed. This code still works, because we never use the non kubernetes branch in production/development. But this does completely break the `!isRunningInsideKubernetes` branch which were present previously.

We should either completely get rid of the usage tracking and only use the direct route. Alternatively, we should change the code such that this is only re-created when the IP has changed.

DanThrane (changes requested):

DanThrane (approved):

PR #3560, opened 2022-07-07
Fixes #3434

### LINKED ISSUE #3434: Projects can have names ending with a space
It is possible to create projects with names that end with whitespace, such as "Project X ".

Bug?

Tiggles:


backend/accounting-service/src/jvmMain/kotlin/services/projects/ProjectService.kt
```diff
@@ -84,6 +84,10 @@ class ProjectService(
         parent: String?,
         principalInvestigatorOverride: String?
     ): String {
+        if (title.trim().length != title.length) {
+            throw RPCException("Project names cannot start or end with whitespace.", HttpStatusCode.BadRequest)
+        }
```
Tiggles, line 89: Should this be in an init-function for the data-class instead?

backend/accounting-service/src/jvmMain/kotlin/services/projects/ProjectService.kt
```diff
@@ -784,6 +788,9 @@ class ProjectService(
         projectId: String,
         newTitle: String
     ) {
+        if (newTitle.trim().length != newTitle.length) {
+            throw RPCException("Project names cannot start or end with whitespace.", HttpStatusCode.BadRequest)
+        }
```
Tiggles, line 793: See above comment

backend/accounting-service/src/jvmMain/kotlin/services/projects/ProjectService.kt
DanThrane: Should be fine here.

DanThrane (approved):

PR #3580, opened 2022-08-09
Product/category parameter for retrieveUtilization

### LINKED ISSUE #2425: [im] retrieveUtilization endpoint should be product specific
Numbers are not easily comparable across different products. This can give a confusing response. If the GPU partition is full and the slim node partition is empty, then the response might indicate a nearly empty cluster. But if the job requires GPU resources then the job might not run any time soon.

xiroV:
Updates `UtilizationService` methods to take a product parameter, returning queue status, capacity and usage of nodes flagged with that product only (instead of the whole cluster). In case no nodes is found, the service will return for the whole cluster (as before).

The call from app-orchestrator have been updated to send the name of the product category.

fixes #2425

### DIFF
--- a/backend/app-kubernetes-service/src/jvmMain/kotlin/services/UtilizationService.kt
+++ b/backend/app-kubernetes-service/src/jvmMain/kotlin/services/UtilizationService.kt
@@ -15,7 +15,7 @@ import kotlinx.serialization.json.JsonPrimitive
 class UtilizationService(
     private val k8: K8Dependencies,
 ) {
-    suspend fun retrieveCapacity(productId: String): CpuAndMemory {
+    suspend fun retrieveCapacity(productCategoryId: String): CpuAndMemory {
         val namespace = k8.client.getResource<Namespace>(
             KubernetesResources.namespaces.withName(NameAllocator.namespace)
         )
@@ -32,7 +32,7 @@ class UtilizationService(
             }
 
         val nodes = allNodes.filter { node ->
-            (node.metadata?.labels?.get("ucloud.dk/machine") as? JsonPrimitive)?.content == productId
+            (node.metadata?.labels?.get("ucloud.dk/machine") as? JsonPrimitive)?.content == productCategoryId
         }.ifEmpty { allNodes }
 
         val nodeAllocatableCpu = nodes.sumOf { node ->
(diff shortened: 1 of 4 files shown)


backend/app-kubernetes-service/src/jvmMain/kotlin/services/UtilizationService.kt
```diff
@@ -15,7 +15,7 @@ import kotlinx.serialization.json.JsonPrimitive
 class UtilizationService(
     private val k8: K8Dependencies,
 ) {
-    suspend fun retrieveCapacity(): CpuAndMemory {
+    suspend fun retrieveCapacity(productId: String): CpuAndMemory {
```
DanThrane, line 18:
```suggestion
    suspend fun retrieveCapacity(productCategoryId: String): CpuAndMemory {
```

backend/app-kubernetes-service/src/jvmMain/kotlin/services/UtilizationService.kt
```diff
@@ -42,25 +46,29 @@ class UtilizationService(
         return CpuAndMemory(nodeAllocatableCpu, nodeAllocatableMemory)
     }
 
-    suspend fun retrieveUsedCapacity(): CpuAndMemory {
-        val jobs = k8.client.listResources<VolcanoJob>(
+    suspend fun retrieveUsedCapacity(productId: String): CpuAndMemory {
```
DanThrane, line 49:
```suggestion
    suspend fun retrieveUsedCapacity(productCategoryId: String): CpuAndMemory {
```

backend/app-kubernetes-service/src/jvmMain/kotlin/services/UtilizationService.kt
```diff
@@ -69,11 +77,15 @@ class UtilizationService(
         return CpuAndMemory(cpuUsage, memoryUsage)
     }
 
-    suspend fun retrieveQueueStatus(): QueueStatus {
-        val jobs = k8.client.listResources<VolcanoJob>(
+    suspend fun retrieveQueueStatus(productId: String): QueueStatus {
```
DanThrane, line 80:
```suggestion
    suspend fun retrieveQueueStatus(productCategoryId: String): QueueStatus {
```

DanThrane (changes requested):
Looks mostly good to me. I would rename a few parameters for clarity.

The frontend also needs to be updated, which is why I am requesting changes.

DanThrane (approved):
Nevermind, I was wrong.

Issue #3611, opened 2022-08-17
Consider merging Icons and Avataaar clothing into single files, respectively


Tiggles:
Changing our bundler to Vite has made it so that every file is served every time the page loads. By bundling into single files, we would hopefully reduce the amount of requests made when loading the page, and also reduce the amount of network requests that occur, that can bring an open DevTool console to its knees.

PR #3614, opened 2022-08-18
File merging 3611

Tiggles:
Reduces file count of about 120.

Merges all icons into one file, merges _some_ avataaar files.


Tiggles:
Fixes #3611 (to an extent)

DanThrane (approved):

PR #3682, opened 2022-09-21
[im] Posix plugin fixes and accounting

### LINKED ISSUE #3661: [im] Accounting with POSIX storage plugin

### LINKED ISSUE #3671: [im] Minor Posix files plugin issues
There's a number of smaller issues related to the Posix plugin:

 - [x] Uploading large files does not work.
 - [x] Copying a file to the same directory will add the increment value in the end of the file name, i.e. `test.txt` -> `test.txt (1)`, but should be `test(1).txt`.
 - [x] Moving a file to a folder (including Trash) which already contains a file with the same name does not work, even with the `WriteConflictPolicy` set to `RENAME`.

xiroV:
fixes #3671

### DIFF
--- a/provider-integration/integration-module/src/main/kotlin/plugins/storage/posix/PosixCollectionPlugin.kt
+++ b/provider-integration/integration-module/src/main/kotlin/plugins/storage/posix/PosixCollectionPlugin.kt
@@ -232,27 +232,7 @@ class PosixCollectionPlugin : FileCollectionPlugin {
 
     private suspend fun calculateUsage(coll: PathConverter.Collection): Long {
         return when (val cfg = pluginConfig.accounting) {
-            "DeviceQuota" -> {
-                TODO()
-                /*
-                memScoped {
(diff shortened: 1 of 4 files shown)


provider-integration/integration-module/src/main/kotlin/plugins/storage/posix/PosixCollectionPlugin.kt
```diff
@@ -180,7 +163,7 @@ class PosixCollectionPlugin : FileCollectionPlugin {
                                 rpcClient
                             ).orThrow()
 
-                            summary.items.forEachGraal inner@{ item ->
+                            summary.items.associateBy { it.id }.values.forEachGraal inner@{ item ->
```
DanThrane, line 166: Breaks graal aot build

DanThrane (commented):
One small fix and it is good to go.

DanThrane (approved):

PR #3707, opened 2022-10-03
Devel im psql

hschu12:
Postgres DB can be given in config. Supports path to embedded or config to external  DB.
All queries has been changed to be in line with postgres syntax
Most of the queries has been tested through frontend and CLI.


provider-integration/integration-module/src/main/kotlin/plugins/storage/ucloud/LimitChecker.kt
```diff
@@ -22,13 +22,14 @@ class LimitChecker(
...
                     """
                         select true
                         from ucloud_storage_quota_locked
                         where
-                            username is :username and
-                            project_id is :project_id and
-                            category is :category
+                            username = :username::text and
+                            project_id = :project_id::text and
+                            category = :category
```
DanThrane, line 32:
```suggestion
                            username is :username::text and
                            project_id is :project_id::text and
                            category is :category
```

DanThrane (approved):
I made a few changes to the embedded postgres. Code looks good now.

PR #3725, opened 2022-10-13
Projectv2 transition

Tiggles:
Transitions to the project v2 API.

Some things that need to be done before merging:
```
    // TODO(Jonas): Handle in general.
    public verifyProject(request: BulkRequest<VerifyProjectRequest>): APICallParameters {
        return apiUpdate(request, this.baseContext, "verifyProject");
    }

```

The `needsVerification` needs to be fetched and added to project on the backend as well.

`ResourcePermissionEditor` is not tested as I could not get the button to appear.


Tiggles:
Known issue: Viewing allocations/suballocations and changing project in context switcher reloads the allocations, but for the wrong project.

Tiggles:
- My Workspace no longer has a dashboard (how to access usage?)
- SharesOutgoing looks at a project at times. Shares should not be available for non-personal workspaces, right?

DanThrane:
> My Workspace no longer has a dashboard (how to access usage?)

It is supposed to work like it did before. If it is gone then we should add it back.

> SharesOutgoing looks at a project at times. Shares should not be available for non-personal workspaces, right?

Correct.

DanThrane (dismissed):
Looks good to me. Let's do a bit of testing on dev before we push it to production.

Tiggles:
Yes, let's please do some testing.

DanThrane (approved):

Issue #1289, opened 2020-02-21
Redesign of application tags


hschu12:
- Better Tag to Application connection. 
  - Requires DB changes.

- Need for predefined list of tags that ADMINs can add to. 
  - Search by typing when adding tags (reduce duplicates).
  - All lowercase.

- Elastic should handle multi-word tags

PR #3727, opened 2022-10-21
Apps backend changes

### LINKED ISSUE #2165: Application tags are case sensitive
Minor issue, but easy to forget since applications tags are always shown in upper case.

xiroV:
Re-implementation of the tag-system for apps. Tags are stored in one database table `tags`, with reference table `application_tags` storing references to `applications` and `tags`. 

Adds an `overview` end-point and database table used to fetch the entire Apps frontpage. The `overview` table can contain references to tools and/or tags (more types of references can be added). The end-point will respond with a ordered list of sections with applications.

The Apps front page should be identical to the old one, with the benefit of only making 2 calls to the backend (one for favorites and one for the remaining sections). Which sections are displayed, their names and order can be managed through the database. Backend end-points does not yet exist for this.

The previous Apps frontpage has not yet been deleted, to allow for easy roll-back.

Does not alter the existing API end-points.

Fixes parts of #1289 with "Search by typing when adding tags (reduce duplicates)" still missing. This should be easy to implement with the new system.

fixes #2165


DanThrane (approved):
Looks good to me. We will need to do a test on dev before we push to production. But we should be able to push this together with the grant update.

PR #3747, opened 2022-11-08
Update react-router

Tiggles:
Initially, this upgrades to the new React-Router version. Whether we actually can upgrade React to v18 with current dependencies.


Tiggles (commented):
I gone through the changes and fixed a few things.

DanThrane (approved):

PR #3750, opened 2022-11-10
Fixes #3749

### LINKED ISSUE #3749: Add a status from status.cloud.sdu.dk in the Support Box
This should let users see if we are aware of an issue with any of our services.

Tiggles:
Shows error is a system is not listed as operational. Removes a bunch of unused code also.
Screenshot 2022-11-10 at 16 09 04


DanThrane (changes requested):
I would change this such that if the status page is down the widget isn't shown at all. Right now it looks like the widget will be shown until a status is loaded (if one is ever loaded).

Tiggles:
Done

DanThrane (approved):

PR #3783, opened 2022-11-30
Move of Syncthing to IM

### LINKED ISSUE #3621: Syncthing sometimes cant restart
When pushing restart on the app (https://cloud.sdu.dk/app/syncthing), Syncthing goes into SUSPENDED state and cannot transition out of it.

Setting it to FAILURE in the DB automatically recreates the syncthing instance for those who are stuck.

### LINKED ISSUE #3659: [sync] Sync should be product-specific
Changes needed for moving the sync feature to the integration module

xiroV:
Fixes bugs related to Syncthing, as well as moved relevant code from app-kubernetes-service to the integration-module. 

It was necessary to make Syncthing product-specific in the process (fixes #3659).
An issue related to restarting Syncthing jobs which sometimes gets stuck in `SUSPENDED` have also been fixed (fixes #3621).

Also solves an issue related to suspended jobs: Suspended jobs could not be cancelled if the job no longer exists in Kubernetes. Only existing jobs could be terminated, which caused a few Syncthing jobs to be locked in a forever-suspended state. A job can now be terminated even if the job does not actually exist in k8 anymore.

A lot of files (particularly markdown) seems to have been touched/removed by api-gen, so please make sure this is correct before merging.

The PR contains changes throughout the stack, but the common functionality have been tested and should work:
 - Add/remove folder, both from Files and Sync UI
 - Add/remove device
 - Restart of Syncthing server
 - Factory reset of Syncthing server

The Files frontend will only show Sync-related functionality if the provider of the current drive/file-collection have a compute-product named `syncthing`.


frontend-web/webclient/app/Files/Files.tsx
```diff
@@ -197,11 +204,28 @@ export const FilesBrowse: React.FunctionComponent<{
...
+    useEffect(() => {
+        if (!syncthingProduct) return;
+        if (didUnmount.current) return;
+        Sync.fetchConfig(syncthingProduct.product.category.provider).then(config => {
             setSyncthingConfig(config);
         });
-    }, []);
+        /*if (!syncthingConfig.loading) {
+            fetchSyncthingConfig(Sync.api.retrieveConfiguration(syncthingProduct.product.category.provider, "syncthing"));
+        }*/
```
Tiggles, line 227:
```suggestion
```

provider-integration/integration-module/src/main/kotlin/plugins/compute/ucloud/Plugin.kt
```diff
@@ -324,6 +325,46 @@ class UCloudComputePlugin : ComputePlugin {
...
     }
+
+    override suspend fun RequestContext.retrieveSyncthingConfiguration(
+        request: IAppsProviderRetrieveConfigRequest<SyncthingConfig>
+    ): IAppsProviderRetrieveConfigResponse<SyncthingConfig> {
+        if (syncthingService == null) {
+            throw RPCException("Not supported by provider", HttpStatusCode.BadRequest)
+        }
+
+        return syncthingService!!.retrieveConfiguration(request)
```
DanThrane, line 336: Avoid `!!`

backend/app-orchestrator-service/api/src/main/kotlin/IntegratedApplications.kt
```diff
@@ -115,7 +115,10 @@ abstract class IntegratedApplicationApi<ConfigType>(
 }
 
 @Serializable
-class IAppsRetrieveConfigRequest<ConfigType>(val providerId: String)
+class IAppsRetrieveConfigRequest<ConfigType>(
+    val provider: String,
+    val product: String,
```
DanThrane, line 120: I think `product` is a bad name in this case. Typically a product is a combination of (name, category, provider). It is not obvious what "product" is in this case.

DanThrane (changes requested):
Some initial comments.

backend/app-orchestrator-service/api/src/main/kotlin/IntegratedApplications.kt
xiroV: Renamed to `productId` since that seems to be the most used (even though it is actually the product name).

DanThrane:
Looks good to me. Although I propose that we wait with merging until after 2022.3.0 is released. Just in case we missed something.

DanThrane:
Seems like we have gotten some conflicts in the mean time. @xiroV Can you fix them? I will approve and merge immediately after.

DanThrane (approved):

Issue #3802, opened 2022-12-07
Timestamp issue for allocations


hschu12:
Running launcher on dev server. 

Backend and Frontend does not seem to agree on timezone.

When making the rootdeposit the start time is set to 00:00::00 UTC on the given date, however when attempting to accept an grant application with the same date failes due to an DB error triggered by the new allocation start date is before the parent start date. 
When the frontend sends the date to the backend the timestamp transaltes to 23:00:00 UTC the day before (22:00:00 if during sommertime)

This might just be an issue when running a dev cluster and only on the first day of running an environment since then the date is always newer than the parent alloc.

hschu12:
Issue fixed by @xiroV in devel accounting merge

PR #3803, opened 2022-12-07
Devel accounting

### LINKED ISSUE #3583: Charge should handle allocations updates in mid/top allocation path
Deal with this situation. Probably want to do this in charge and make sure we don't go above the  initialBalance

         Initial state:
         1000
         500
         100

         Charge 50 in leaf
         1000 (950)
         500 (450)
         100 (50)

        Update leaf to have 10 only
         10 (-40)
         500 (450)
         100 (50)

         Charge 0 in leaf
         10 (60)
         500 (500)
         100 (50)

### LINKED ISSUE #3584: Synchronization with postgres should not remove IsDirty before update has taken place.
Currently we set isDirty to false before we push to Postgres.
This results in a risk of update to Postgres going wrong and our in-memory thinking it has been pushed.

### LINKED ISSUE #3585: Handle Granted in val in sync

### LINKED ISSUE #3586: Write Test for Gifts system

### LINKED ISSUE #3587: Write tests for Grant system

### LINKED ISSUE #3589: Implement synchronization of Grants/Gifts approval
Grants risk being approved and then lose the entire allocation if we miss the deposit. 
We need a way of ensuring that a when grant is approved the ressources are also allocated.

One way is to mark a grant as fully
completed once we have finished synchronizing the data. The same holds true for gifts.

### LINKED ISSUE #3590: Implement distributed lock
Only one instance of Accoutning service should be allowed to handle the processing of accounting to ensure data consitency

### LINKED ISSUE #3591: Migrate old Postgres functions
Need to remove the old charge, deposit etc functions from the postgres code.

### LINKED ISSUE #3592: Make sure requests are sent to the processing instance
When running mulitiple accounting services we need to know where to forward the requests to.

See auth.useriterationservice for procedure.

### LINKED ISSUE #3593: Make tests for deposit, charge and update of funds

### LINKED ISSUE #3594: Implement filter out empty allocations
When choosing an allocation to a subproject it should not suggest to use an already empty allocation.
This is the cause of many support tickets.

Ideally it will also sort on balance so that highest available balance is at the top of suggestions.

This can be done through request flags

### LINKED ISSUE #3595: Implement maxUsableBalance val
When retrieving and wallet or allocation, then also calculate what is the maximum amount of resources actually avalable.
Eventhough you have been given an allocation of 1000DKK then the parent project might only have 200 DKK available through that particullar allocation. This results in confusion because the user might think they have 800 left when the frontend starts complaining that there are not enough resources and it is in fact the parent project that are out of resources.

### LINKED ISSUE #3605: Implement Retrieve Wallets

### LINKED ISSUE #3643: In memory db ruins first time setup in launcher
First time we start launcher we create wallets, allocations, user and products.
Only the products and users are in DB after the setup of UCloud.
The DB does not manage to sync to DB before the process is killed.

### LINKED ISSUE #3660: Browse suballocations should paginate with processor.
Does not get updated values back if using Postgres

### LINKED ISSUE #3724: Stress test Accounting charge

### LINKED ISSUE #3736: Problem with multiple syncs to DB when running multiple requests
When multiple requests are hitting the processor in a short time (time shorter than a synchronization to DB) it triggers multiple syncs.

Next sync is only set after sync is done and therefore the check beforehand grants a sync for each request recieved

### LINKED ISSUE #3737: Accounting caches are not clean when testing
When performing tests, the allocations, wallets and transactions are not cleaned between tests, but DB is. Resulting in projects and other things not being available for wallets that later are attempted to become syncronized

### LINKED ISSUE #3772: Introduce new properties on projects related to accounting
# New `Project` property: `canConsumeResources`

This new property allows us to control the behavior for owners which should not be able to consume resources, typically, because they are only supposed to further sub-allocate the resources. This information will be returned as part of the project API. 

The frontend will use this information to display an error message when attempting to browse/create resources.

The resource system will use this to enforce that you cannot create any resources.

The accounting system will ignore this property.

This value will be true for all existing project. Some projects will manually have this value changed in production. We will coordinate with the projects to make sure that they have no data stored in these projects. This value is immutable and cannot be changed after creation. Initially, it will not be possible to change this through the API/UI.

The backend will pass this information to the providers. Providers can use this information to ignore certain projects.

# New `WalletAllocation` property: `canAllocate` and `allowSubAllocationsToAllocate`

The `canAllocate` property controls if a given allocation is capable of being sub-allocated. Sub-allocations occur with every deposit/transfer. If this is false, then deposit and transfers with this allocation will always fail.

All allocations, except for root allocations, have this value set based on their parent's `allowSubAllocationsToAllocate` property. The `allowSubAllocationsToAllocate` property is also inherited from parents. If the wallet owner is a personal workspace, then the value always becomes false. It will not be possible to set this value from the UI to begin with.

Both properties are immutable.

Allocations which are owned by personal workspaces will have their `canAllocate` set to false. Projects will be true.

The frontend will use this value to control if it should be possible to use a given allocation for a sub-allocation.

The backend accounting system will enforce these values.

hschu12:
There is still is a problem with timestamps on new grant application (#3802 ) but this might just be a launcher issue.

Fixes #3583 
Fixes #3584 
Fixes #3585 
Fixes #3586 
Fixes #3587 
Fixes #3589 
Fixes #3590 
Fixes #3591
Fixes #3592  
Fixes #3593
Fixes #3594 
Fixes #3595 

Fixes #3605
Fixes #3643 
Fixes #3660 

Fixes #3724 
Fixes #3736
Fixes #3737 
Fixes #3772  (already closed on master)


backend/auth-service/src/main/kotlin/http/PasswordController.kt
```diff
@@ -28,6 +28,8 @@ class PasswordController(
                     throw RPCException.fromStatusCode(HttpStatusCode.BadRequest)
                 }
 
+                println(params)
```
Tiggles, line 31: Remove

backend/service-lib/src/main/kotlin/calls/ApiConventions.kt
```diff
@@ -602,7 +602,7 @@ verification API to cleanup these resources later.
 @UCloudApiStable
 data class BulkRequest<out T : Any>(val items: List<T>) {
     init {
-        if (items.size > 1_000 || items.isEmpty()) throw RPCException("BulkRequest must not be empty", HttpStatusCode.BadRequest)
+        if (items.size > 1_000 || items.isEmpty()) throw RPCException("BulkRequest must not be empty or exceed limit", HttpStatusCode.BadRequest)
```
Tiggles, line 605: This will be shown as an error message on the frontend, right? This is not a great error message to show the user. With or with the "or exceed limit" part.

Tiggles: Hopefully, this would only occur when doing development, but if it gets through, the user won't know what to do different.

frontend-web/webclient/app/Admin/Overview.tsx
```diff
@@ -16,6 +16,7 @@ const linkInfo: LinkInfo[] = [
     {to: "/admin/appk8", text: "Maintenance (UCloud)", icon: "calendar", color: "white", color2: "black"},
     {to: "/admin/providers", text: "Providers", icon: "cloudTryingItsBest", color: "white", color2: "white"},
     {to: "/admin/scripts", text: "Scripts", icon: "play", color: "white", color2: "white"},
+    {to: "/admin/devData", text: "Development Test Data", icon: "activity", color: "white", color2: "white"},
 ];
```
Tiggles, line 21:
```suggestion
];

if (DEVELOPMENT_ENVIROMENT) {
    linkInfo.push({to: "/admin/devData", text: "Development Test Data", icon: "activity", color: "white", color2: "white"}),
}

```

Tiggles: I believe `devData` only makes sense on the `dev` environment, so I wouldn't allow it elsewhere, if that is the case.

Note: The predicate is not correct in this suggestion.

frontend-web/webclient/app/Project/Allocations.tsx
```diff
@@ -326,6 +325,27 @@ const WalletViewer: React.FunctionComponent<{wallet: Wallet}> = ({wallet}) => {
     </>
 }
 
+function availableBalance(allocation: WalletAllocation, wallet: Wallet) :JSX.Element {
```
Tiggles, line 328:
```suggestion
function AvailableBalance(props: {allocation: WalletAllocation, wallet: Wallet}) :JSX.Element {
```

frontend-web/webclient/app/Project/Allocations.tsx
```diff
@@ -345,6 +365,7 @@ export const AllocationViewer: React.FunctionComponent<{
                     <Box flexGrow={1} />
                 </Flex>}
                 <div>{usageExplainer(allocation.initialBalance - allocation.balance, wallet.productType, wallet.chargeType, wallet.unit)} used</div>
+                {availableBalance(allocation, wallet)}
```
Tiggles, line 368:
```suggestion
                <AvailableBalance allocation={allocation} wallet={wallet} />
```

frontend-web/webclient/app/Project/ContextSwitcher.tsx
```diff
@@ -122,12 +122,17 @@ function onProjectUpdated(navigate: NavigateFunction, runThisFunction: () => voi
     const {pathname} = window.location;
     runThisFunction();
     const splitPath = pathname.split("/").filter(it => it);
+    console.log(splitPath);
```
Tiggles, line 125: Remove

Tiggles (dismissed):
Nothing blocking, but I have some things I think can be improve before merging.

DanThrane:
Tests passed, but slightly out of date due to frontend change. I am merging regardless.

PR #3886, opened 2023-01-30
Update Main.kt

schulztest:
I think it would be nice to print RYUNNING for this test


backend/accounting-service/src/main/kotlin/Main.kt
```diff
@@ -34,5 +34,6 @@ data class Configuration(
 )
 
 fun main(args: Array<String>) {
+    println("RUNNING")
```
DanThrane, line 37:
```suggestion
    println("RYUNNING")
```

PR #3894, opened 2023-01-31
Project link invites

### LINKED ISSUE #3421: Allow invite to project via link
PIs and admins of a project should be able to create an invitation link to a project. This link should be accessible by any UCloud user, once used, the authenticated user must automatically be added to the project.

This feature depends on being able to pass a "redirection" page, see #3420.

xiroV:
fixes #3421 

Implemented as described in the issue/design document.


backend/accounting-service/src/main/kotlin/services/projects/v2/ProjectService.kt
```diff
@@ -1245,6 +1249,267 @@ class ProjectService(
...
+                    select * from project.invite_links
+                    left join project.invite_link_group_assignments a on a.link_token = token
+                    left join project.groups g on g.id = a.group_id
+                    where project_id = :project and now() < expires
+                    order by expires desc
+                """
+            ).rows.groupBy { it.getAs<UUID>("token") }.map { entry ->
+                ProjectInviteLink(
+                    entry.key.toString(),
+                    entry.value.first().getAs<OffsetDateTime>("expires").toEpochSecond(),
```
DanThrane, line 1315: All other timestamps in UCloud uses epoch millis.

backend/accounting-service/src/main/kotlin/services/projects/v2/ProjectService.kt
```diff
@@ -1245,6 +1249,267 @@ class ProjectService(
...
+
+        val result = ctx.withSession { session ->
+            requireAdmin(actorAndProject.actor, listOf(project), session)
+
+            session.sendPreparedStatement(
+                {
+                    setParameter("project", project)
+                },
+                """
+                    select * from project.invite_links
```
DanThrane, line 1306: Would probably be better if this selects the required columns instead of selecting everything.

backend/accounting-service/src/main/kotlin/services/projects/v2/ProjectService.kt
```diff
@@ -1245,6 +1249,267 @@ class ProjectService(
...
+                    update project.invite_links
+                    set role_assignment = :role
+                    where project_id = :project and token = :token
+                """
+            ).rowsAffected > 0
+
+            if (!success) {
+                throw RPCException(
+                    "Unable to update role assignment",
+                    HttpStatusCode.BadRequest
```
DanThrane, line 1384: A better error message should probably state that the link might no longer be valid/has been deleted.

backend/accounting-service/src/main/kotlin/services/projects/v2/ProjectService.kt
```diff
@@ -1245,6 +1249,267 @@ class ProjectService(
...
+            requireAdmin(actorAndProject.actor, listOf(project), session)
+
+            session.sendPreparedStatement(
+                {
+                    setParameter("project", project)
+                    setParameter("token", request.token)
+                    setParameter("groups", request.groups)
+                },
+                """
+                    with new (group_id, token) as (
```
DanThrane, line 1407: I would add more newlines here to be consistent with other SQL queries in UCloud/Core.

backend/accounting-service/src/main/kotlin/services/projects/v2/ProjectService.kt
```diff
@@ -1245,6 +1249,267 @@ class ProjectService(
...
+                        select id, cast(:token as uuid) from project.groups where project = :project and id in (select unnest(cast(:groups as text[])))
+                    ), changed as (
+                        insert into project.invite_link_group_assignments (group_id, link_token)
+                        select * from new
```
DanThrane, line 1411: I would avoid using `select *` as input to an `insert` statement. This seems like something that is very easy to accidentally break.

backend/accounting-service/src/main/kotlin/services/projects/v2/ProjectService.kt
```diff
@@ -1245,6 +1249,267 @@ class ProjectService(
...
+                        link_token = cast(:token as uuid)
+                    where
+                        m.role = 'PI';
+                """
+            ).rows.groupBy { it.getString("project") }.map { entry ->
+                val project = entry.key
+                val pi = entry.value.first().getString("pi")
+                val role = entry.value.first().getString("role")
+
+                if (project == null || pi == null || role == null) {
```
DanThrane, line 1453: This will never happen since `first()` will crash the request with an internal server error instead. See https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/first.html#exceptions

DanThrane (changes requested):

Tiggles:
This will merge into the wrong branch.

backend/accounting-service/src/main/kotlin/services/projects/v2/ProjectService.kt
```diff
@@ -1245,6 +1248,272 @@ class ProjectService(
...
+                )
+            }.firstOrNull() ?: throw RPCException("Link expired", HttpStatusCode.BadRequest)
+
+            createInvite(
+                projectAssignment.piActorAndProject,
+                bulkRequestOf(ProjectsCreateInviteRequestItem(actorAndProject.actor.safeUsername())),
+                ctx
+            )
+
+            acceptInvite(
```
DanThrane, line 1477: As far as I can tell, this has the unfortunate side-effect of sending an email to the user that they have been invited to the project. We should obviously not send such an email. I suggest we add a new parameter to `createInvite` and `changeRole` which disables notifications and emails.

backend/accounting-service/src/main/kotlin/services/projects/v2/ProjectService.kt
```diff
@@ -1245,6 +1248,272 @@ class ProjectService(
...
+                    bulkRequestOf(
+                        projectAssignment.groups.map {
+                            GroupMember(actorAndProject.actor.safeUsername(), it)
+                        }
+                    ),
+                    ctx
+                )
+            }
+
+            changeRole(
```
DanThrane, line 1495: See email comment above.

DanThrane (changes requested):
Seems like we forgot about emails and notifications. See review comments for details.

DanThrane (approved):
Looks good to me.

PR #3895, opened 2023-02-01
Jenkins setup

hschu12:
Setup done 
Changes made to createUser in Auth to make illegal chars fail on creation. Test now succeeds
Also contains update for k8 file for elastic-management which was forgotten in last PR.
Only slurm tests fail now


DanThrane (approved):
Looks good to me. I am merging this directly with `master` since there is no point in going through `staging` for this one. I am also merging despite the fact that the Slurm tests are failing, since we know that the testing suite is still not working reliably for this sub-system.

PR #3896, opened 2023-02-02
Elastic cleanup fix

hschu12:
Error when using * as index.
Resulted in also getting system indices such as .security, .kibana etc. 
These indices has not been updated to version 8, so I guess that why they are missing type property in json.

Fix is to go for all http_logs*, infrasturcture* and kubernets* indstead where we used * before.

Elastic cleanup has been redeployed on production with fix and seems to run just fine.


DanThrane (approved):
Looks good to me. Merging directly with `master` since the bug fix is fixing a broken production service.

DanThrane (approved):

PR #3904, opened 2023-02-08
Jenkinshotfix

hschu12:
Added buildstep to catch compile errors
Added more info on jenkins alert


hschu12:
Runs as intented on original branch and on PR

DanThrane (approved):

PR #3912, opened 2023-02-16
Increase height of Provider cards.

Tiggles:
Add 'Show all' to Provider-card
Add DTU to login page
Add dtu-logo to Assets. Already in /public/ but I think this could be a bit fragile.


DanThrane (approved):
I am pushing this directly to `master` since the PR is urgent.

PR #3917, opened 2023-02-21
File-sharing with links

### LINKED ISSUE #3900: Allow sharing folders with link
In relation to #3421 we discussed adding the same feature to Shares, making it possible for users to create shares (share access to files) by sharing a created link.

Shares can currently only be created if the sharer knows the username of the recipient. The ability to share with a link instead would resolve this problem.

It is possible that much of the code from #3894 can be reused. My initial thoughts on what needs to be changed:

**Back-end**
   - Since automatic assignment to "groups" is irrelevant, only one database table is needed and also eliminates the need for one end-point, `updateInviteLinkGroupAssignment`.
   - Automatic role-assignment can be changed to permission-assignment. Default should be `READ`.
   - A file-path to the share will be stored instead of a project ID. It is important to check if the file-path still exists on `accept`.
   - File operations (delete, move, etc.) should delete the invite link as well.
   - `accept` should create a new share on behalf of the sharer, and automatically approve by the client. 
 
**Front-end**
  - UI can generally be reused, but it is not obvious where it should be placed:
    - In the current "Share folder" dialog, under a separate tab.
    - In its own dialog "Share with link" (add a new operation to Files).
    - Other ideas are welcome.

xiroV:
fixes #3900 

Implemented in a similar way as Project invite links (described in #3421)


backend/file-orchestrator-service/api/src/main/kotlin/Core.kt
```diff
@@ -518,4 +518,6 @@ data class FSFileSupport(
 
     @UCloudApiDoc("Declares support for the streamingSearch endpoint")
     val streamingSearchSupported: Boolean = false,
+
+    val sharesSupported: Boolean = true
```
DanThrane, line 522: At this point in time, it would be a more safe bet to default to false. We risk accidentally enabling this on Hippo.

backend/file-orchestrator-service/src/main/kotlin/service/ShareService.kt
```diff
@@ -667,4 +704,291 @@ class ShareService(
...
+
+        return ctx.withSession { session ->
+            val count = session.sendPreparedStatement(
+                {
+                    setParameter("file_path", request.path)
+                },
+                """
+                    select count(*) from file_orchestrator.shares_links where file_path = :file_path and now() < expires
+                """
+            ).rows.first().getLong("count") ?: 0
```
DanThrane, line 737:
```suggestion
            ).rows.first().getLong(0) ?: 0
```

DanThrane (changes requested):
Most of this looks good to me, but I have a few suggested changes.

DanThrane (approved):

PR #4037, opened 2023-07-12
Load simulation

### LINKED ISSUE #4009: Load simulation
Issues related to high amounts of traffic, and regular traffic over long periods of time, are difficult for us to debug.

These issues are often not reproducible on the development system due to minimal traffic, while our local development environments are restarted so often that we often don't catch side-effect of long-running processes.

By simulating user traffic over longer periods of time on the development system, we might be able to catch some of these issues before they reach the production system.

xiroV:
Will simulate requests from a specified number of "fake" users, to test how the system behaves under load.

fixes #4009


DanThrane (approved):
lgtm

PR #4042, opened 2023-07-26
fixes #4016. Problem when alloc is null

### LINKED ISSUE #4016: Allocation update call never returns
Observed issue:

- Open the user interface for resource allocations
- Make an update to one of the allocations
- The backend never responds
- But the update is registered by the backend and visible after a refresh of the page

hschu12:
fixes #4016. Problem when alloc is null it crashes silently.
Also a small fix for launcher that did not start correctly due to auto created user did not have valid email.


hschu12:
Fix is deployed on production

backend/accounting-service/src/main/kotlin/services/wallets/AccountingProcessor.kt
```diff
@@ -1273,19 +1273,23 @@ class AccountingProcessor(
...
-            val alloc = allocations[i]!!
-            val newNotAfter =
-                if (min(alloc.notAfter ?: Long.MAX_VALUE, parent.notAfter ?: Long.MAX_VALUE) == Long.MAX_VALUE) {
-                    null
-                } else {
-                    min(alloc.notAfter ?: Long.MAX_VALUE, parent.notAfter ?: Long.MAX_VALUE)
+            //NOTE(Henrik) It is possible to have nullable allocations in the list
+            //In that case continue to next id
+            val alloc = allocations[i]
+            if (alloc != null) {
```
Tiggles, line 1279: I don't know if this reads better, but it believe could be
`val alloc = allocations[i] ?: continue

DanThrane (approved):

PR #4043, opened 2023-08-02
simple du on collections for weka scan

### LINKED ISSUE #3878: Project administrators cannot see the created invitations of other project administrators
Imagine a project P with two admins A1 and A2 they both wish to invite the user B. The following are the steps to reproduce:

- A1 invites B to P
- B does not respond immediately
- A2 opens the page to see invites. They do not see any.
- A2 invites B to P
- A2 receives an error because the invitations could not be created (because one already exists)

This issue is caused by the slightly too primitive filter which is p resent in `browseInvites`. When viewing outgoing invites it check only for invites that are created by the authenticated user. Which means that A2 is unable to see the invitations created by A1.

### LINKED ISSUE #3994: UCloud storage plugin fails silently when moving files between systems
This is known not to work, since it is not possible to move files between filesystems. This is intended behavior. However, this should not fail silently. Instead an error message should be displayed to the error describing the failure.

### LINKED ISSUE #4041: Update storage scan to handle Weka system
Currently we only scan our ceph system. Since we are moving old data to Weka and creating all new drives on this, we should make the scan for Weka ASAP.

hschu12:
Made a small scan using "du -s path" command from scan module.
Martin thinks Weka can handle this fast, but we might want to parallize it if it just seems slow and not totally useless.

Tested this on local branch env vs a ceph storage by only having else branch. 

Should add WekaFS as enum but not sure what or if we have it as an type in DB yet?

Fixes #4041 
Fixes #3878 
Fixes #3994


provider-integration/integration-module/src/main/kotlin/plugins/storage/ucloud/FastDirectoryStats.kt
```diff
@@ -38,6 +42,25 @@ class CephFsFastDirectoryStats(private val nativeFs: NativeFS) : Loggable, FastD
     }
 }
 
+class WekaDirectoryStats(private val nativeFs: NativeFS) : Loggable, FastDirectoryStatsInterface {
+    override val log = logger()
+
+    override suspend fun getRecursiveSize(file: InternalFile, allowSlowPath: Boolean): Long? {
+        return try {
```
DanThrane, line 49: Should return `null` when `allowSlowPath = false`.

provider-integration/integration-module/src/main/kotlin/plugins/storage/ucloud/FastDirectoryStats.kt
```diff
@@ -38,6 +42,25 @@ class CephFsFastDirectoryStats(private val nativeFs: NativeFS) : Loggable, FastD
...
+            val (_, stdout, _) = executeCommandToText("/usr/bin/du") {
+                addArg("-s")
+                addArg(file.path)
+            }
+            val regex = "^\\d+".toRegex()
```
DanThrane, line 54: We can probably do something more efficient than a regex here.

DanThrane (changes requested):

provider-integration/integration-module/src/main/kotlin/plugins/storage/ucloud/FastDirectoryStats.kt
```diff
@@ -38,6 +42,28 @@ class CephFsFastDirectoryStats(private val nativeFs: NativeFS) : Loggable, FastD
...
+class WekaDirectoryStats(private val nativeFs: NativeFS) : Loggable, FastDirectoryStatsInterface {
+    override val log = logger()
+
+    override suspend fun getRecursiveSize(file: InternalFile, allowSlowPath: Boolean): Long? {
+        if (!allowSlowPath) {
+            return null
+        }
+        return try {
+            val (_, stdout, _) = executeCommandToText("/usr/bin/du") {
+                addArg("-s")
```
DanThrane, line 54: Use `-sb` (comment from Martin)

DanThrane (changes requested):

PR #4067, opened 2023-10-27
Improvements to file preview

### LINKED ISSUE #4066: Improve file preview

xiroV:
Improves preview of files, with the two main changes:

 - Use magic bytes ([magic-bytes.js](https://github.com/LarsKoelpin/magic-bytes)) to determine the file type, instead of the file extension. Preview of files should now work in most cases, independently of the extension. 
 - Switch to [monaco-editor](https://microsoft.github.io/monaco-editor/) for preview of text (set to read-only mode). The editor includes more features like file-search and better performance. The 100kb limit on syntax highlighting has been removed since it seems monaco has build-in limits. Monaco is not able to determine the file type, so we try to guess the file type from the extension and pass it to monaco.

fixes #4066


DanThrane (approved):
Looks good to me

PR #4073, opened 2023-11-01
Fix #3607 and also extract data for deic update

### LINKED ISSUE #3607: `Member files:` and `Home` drive resolves to same mount path when mounted to job
Related to #3184

When drives/filecollections are mounted to an app, it will be mounted to a folder using the id of the drive, as described in the above issue.

However both `Member files: user` and the `Home` drive of `user` will mount to a folder named `user` as far as I can see.

Attempting to start a job on dev with both `Member files: user` and the `Home` drive of user mounted seems to lead to the job never leaving the queue.

hschu12:


DanThrane (approved):

provider-integration/integration-module/src/main/kotlin/plugins/compute/ucloud/FeatureFileMount.kt
```diff
@@ -135,10 +135,21 @@ class FeatureFileMount(
         )
 
         fileMounts.forEach { mount ->
+            //TODO(HENRIK) Is there a better way to check if it a drive??
```
DanThrane, line 138: Nevermind me initial approval, let's check if we can actually improve this bit.

DanThrane (changes requested):

provider-integration/integration-module/src/main/kotlin/plugins/compute/ucloud/FeatureFileMount.kt
DanThrane: I will check later today.

DanThrane (approved):

PR #4093, opened 2023-11-23
fixes corehour charge error

hschu12:


launcher-tool/src/main/kotlin/ComposeFile.kt
```diff
@@ -511,6 +511,14 @@ sealed class ComposeService {
                           memory: 1
                           description: An example CPU machine with 1 vCPU.
                           pricePerHour: 0.5
+                      cpu-h:
+                        cost: 
+                          type: Resource 
+                          interval: Minutely
```
Tiggles, line 517: Is 'Minutely' used generally? Because it seems like a weird word to use in this context.

DanThrane (approved):

PR #4103, opened 2023-12-07
Mount shared

### LINKED ISSUE #3965: Unable to mount root of a shared folder
Steps to reproduce:

- Create a share
- Mount the share itself (by selecting "Use this folder")
- Observe that the folder is not mounted

Note that this only happens when you use the root of the mount, it does not occur when you select any subfolder of the share.

hschu12:
Added support for mounting shares as root.  Fixes #3965 . 
Also a small change to mount path to better reflect frontend names for Home and Member Files.


backend/app-orchestrator-service/src/main/kotlin/services/JobVerificationService.kt
```diff
@@ -81,7 +87,9 @@ class JobVerificationService(
         // Check mounts
         run {
             val mounts = resources.filterIsInstance<AppParameterValue.File>()
+            println(mounts)
             newResources.addAll(checkAndReturnValidFiles(actorAndProject, mounts))
+            println(newResources)
```
Tiggles, line 92:
```suggestion
```

backend/app-orchestrator-service/src/main/kotlin/services/JobVerificationService.kt
```diff
@@ -126,26 +132,63 @@ class JobVerificationService(
...
             it.path.split("/").last()
         }.toSet()
-        if (files.size != filenames.size) {
+        if (translatedFiles.size != filenames.size) {
             throw RPCException("Cannot mount files with same name", HttpStatusCode.BadRequest)
         }
 
-        for (file in files) {
+        for (file in translatedFiles) {
+            println("file: $file")
```
Tiggles, line 191:
```suggestion
```

DanThrane:
I have pushed an additional commit. It should take of the bug related to mounting the root of non-shares, but I haven't tested it yet. The arg-builder bug still needs to be resolved.

backend/file-orchestrator-service/api/build.gradle.kts
```diff
@@ -1,4 +1,5 @@
 dependencies {
    api(project(":accounting-service:api"))
-   // api("com.github.java-json-tools:json-schema-validator:2.2.14")
+    implementation(project(mapOf("path" to ":service-lib-server")))
```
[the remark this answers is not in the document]
hschu12: Automatic import when tries to use metadata service. Forgot to remove.

provider-integration/integration-module/src/main/kotlin/plugins/compute/ucloud/FeatureParameter.kt
```diff
@@ -127,7 +127,13 @@ private class OurArgBuilder(
                     if (components.isEmpty()) {
                         return ArgumentBuilder.Default.build(parameter, value)
                     }
-                    joinPath("/work", components[components.lastIndex]).removeSuffix("/")
+                    if (components.size == 4 && components[2] == "home") {
```
DanThrane, line 130:
```suggestion
                    if (components.size == 4 && components[2] == PathConverter.HOME_DIRECTORY) {
```

provider-integration/integration-module/src/main/kotlin/plugins/compute/ucloud/FeatureParameter.kt
```diff
@@ -127,7 +127,13 @@ private class OurArgBuilder(
...
+                        joinPath("/work", "Home")
+                    } else if (components.size == 6 && components[4] == "Members' Files"){
```
DanThrane, line 132:
```suggestion
                    } else if (components.size == 6 && components[4] == PathConverter.PERSONAL_REPOSITORY){
```

DanThrane (changes requested):
Minor changes

DanThrane (approved):

PR #4107, opened 2023-12-14
added check to Ips on jobCreation

### LINKED ISSUE #4104: Same IPs can be used at multiple jobs
Should not be possible to use the same IP resource with multiple jobs at the same time

hschu12:
fixes #4104


DanThrane (dismissed):
I will approve this, but it still doesn't fix the issue which is present in UCloud/Core (this shouldn't even be getting to the provider).

DanThrane (approved):

Issue #4096, opened 2023-11-28
[ui-redesign] App store issues


xiroV:
Issues related to app store found after latest deployment to dev.

- [x] Flavor name is missing in app studio.
- [x] Changing flavor on Create Job page sets the version as well, but not to newest version. Should probably not set version.
- [x] Title (as in `<title>`) on Create Job page displays `null` if no version is selected.
- [ ] Database occasionally locks? Is unable to recover from this state. Needs further investigation.

PR #4109, opened 2023-12-18
Various fixes and job event notifications

### LINKED ISSUE #4088: Send an e-mail when job starts, if time-threshold is exceeded
Would require that are user opts in to the feature. Adding a checkbox on the runs page would solve this.
Who would receive the e-mail, anyone in the project, or only the person who launched the app?
What should the threshold be? 5 minutes?

### LINKED ISSUE #4089: Show a front-end notification when a job transitions to RUNNING
Always show, no matter the time it took.
But maybe don't show if already viewing the job.
No requirements for opting in.
Show across all projects?
Show on for person who started the job, or every job across all projects?
Only show for active project?
Play a sound on transition?

### LINKED ISSUE #4106: EventController exceptions due to missing project plugin for k8 provider

xiroV:
Fixes various smaller bugs.

 - Notifications send from the backend now works without having to refresh the page (hasn't been working for a long time due to a small bug).
 - Send notification to user when a job: starts, completes, fails or expires (fixes #4089)
 - Added options for mail-notifications for job events: starts or stops (due to success, failure or expiration). These are disabled by default and has to be enabled manually by the user. Fixes #4088 (although without the threshold requirement).
 - Browser title should be correct on the create-job page now.
 - Show current flavor in the flavor-input field in the Application studio.
 - fixes #4106.
 - Changing flavor on create-job page should now ignore the version in the call, thus always selecting the newest version of the application.

fixes (most of) #4096


frontend-web/webclient/app/Notifications/index.tsx
```diff
@@ -58,16 +58,42 @@ function resolveNotification(event: Notification): {
...
 
 function onNotificationAction(notification: Notification, navigate: NavigateFunction) {
     switch (notification.type) {
-        case "APP_COMPLETE":
-            navigate(`/applications/results/${notification.meta.jobId}`);
+        case "JOB_COMPLETED":
+            navigate(`/jobs/properties/${notification.meta.jobId}`);
+            break;
+        case "JOB_STARTED":
+            navigate(`/jobs/properties/${notification.meta.jobId}`);
```
Tiggles, line 96: What about expired/failed?

xiroV:
I've updated the PR according to our discussion at our last meeting.

Notifications and emails related to jobs will now be added to a list instead of being sent immediately. Any additional requests that will trigger a notification/email of the same type will be added to the same list. After 10 seconds (from the first request) a notification (and email, if wanted by the user) will be sent.

The message sent to the user depends on the number of notifications of that type, i.e. for 10 jobs and 1 job respectively:

> 10 jobs completed successfully.

> Your Coder C++ job completed successfully.

The exact wording of these notifications and mail templates might need to be adjusted.

peter-sk:
Awesome. Looking forward to this.

backend/app-orchestrator-service/src/main/kotlin/services/JobResourceService.kt
```diff
@@ -474,6 +518,125 @@ class JobResourceService {
...
+                        if (it.jobName != null) {
+                            "${it.jobName} (${it.jobId})"
+                        } else {
+                            it.jobId
+                        }
+                    }
+
+                    val appTitles = sameType.map { it.appTitle }
+
+                    when (notification.type) {
```
DanThrane, line 612: I'm probably nitpicking, but I think it would be useful if we could bundle different states into a single email as well.

For example, assume these events occur within the "collection window":

- Job 41231 start
- Job 41232 start
- Job 41233 start
- Job 41231 fail
- Job 41000 completes
- Job 41232 fail

Would result in a single email:

Subject: Your job(s) on UCloud has started
Body:
3 of your jobs have started and 3 of your jobs have completed.

- [LINK TO JOB: 41231]: ${jobTitle ?: applicationTitle} has started
- [LINK TO JOB: 41232]: ${jobTitle ?: applicationTitle} has started
- [LINK TO JOB: 41233]: ${jobTitle ?: applicationTitle} has started
- [LINK TO JOB: 41231]: ${jobTitle ?: applicationTitle} has completed (failure)
- [LINK TO JOB: 41000]: ${jobTitle ?: applicationTitle} has completed
- [LINK TO JOB: 41232]: ${jobTitle ?: applicationTitle} has failed

DanThrane (changes requested):
We need to fix the concurrency issue and I think we can do slightly better for the email format.

frontend-web/webclient/app/UserSettings/ChangeNotificationSettings.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+export const ChangeNotificationSettings: React.FunctionComponent<{setLoading: (loading: boolean) => void}> = () => {
+    const [commandLoading, invokeCommand] = useCloudCommand();
+    const [state, dispatch] = React.useReducer(reducer, initialState, () => initialState);
+    const info = useCallback(async () => {
+
+        const notificationSettings = await invokeCommand(
+            retrieveNotificationSettings({}),
+            { defaultErrorHandler: false }
+        );
```
Tiggles, line 46: If you're using `defaultErrorHandler: false`, shouldn't you wrap this in try-catch?

DanThrane (approved):

PR #4113, opened 2023-12-21
fixes #4092 . Also removes single warning and updates gradle wrapper

### LINKED ISSUE #4092: Deleting a drive does not delete the entry in the UCloud provider plugin
Steps to reproduce:

- Create a project
- Start a job, such that member files are initialized
- Stop the job
- Delete the member files drive
- Start a new job after waiting for one minute (wait for cache to be invalidated)
- This results in a "Not found"

The `ucloud_storage_drives` table still contains the old row, which it should not.

hschu12:


provider-integration/integration-module/src/main/kotlin/plugins/storage/ucloud/DriveLocator.kt
```diff
@@ -288,6 +288,33 @@ private object DriveAndSystemStore {
         }
     }
 
+    suspend fun delete(drive: UCloudDrive, ctx: DBContext = dbConnection) {
+        val entry = entries.indexOfFirst { it.drive.ucloudId == drive.ucloudId }
```
DanThrane, line 292: Mutex should be around this part and not the database part.

DanThrane (changes requested):
Minor change required

DanThrane (approved):

PR #4122, opened 2024-01-15
fixes #3898

### LINKED ISSUE #3898: No feedback when creating folder that already exists
The backend returns 200 when the folder already exists resulting in no feedback to the user when attempting to create it.

hschu12:
Now finds suitable name on rename policy when creating folders and creates a folder with that name instead.
filename(X) 
X=1..1000

### DIFF
--- a/provider-integration/integration-module/src/main/kotlin/plugins/storage/ucloud/FileQueries.kt
+++ b/provider-integration/integration-module/src/main/kotlin/plugins/storage/ucloud/FileQueries.kt
@@ -131,6 +131,21 @@ class FileQueries(
         }
     }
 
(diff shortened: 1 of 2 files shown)


DanThrane (approved):

provider-integration/integration-module/src/main/kotlin/plugins/storage/ucloud/FileQueries.kt
```diff
@@ -131,6 +131,21 @@ class FileQueries(
         }
     }
 
+    suspend fun findAvailableNameOnRename(id: String): String {
+        for (i in 1..1000) {
```
DanThrane, line 135:
```suggestion
        for (i in 1..10000) {
```

Issue #4120, opened 2024-01-15
Issues with file upload conflict policy


xiroV:
Some minor issues with file upload have been reported to Mads by a user. After a short investigation it seeems that:

 - The backend is not respecting the conflict policy defined by the frontend,
 - If the conflict policy is set to `REPLACE`, the existing file is not truncated properly.

Tiggles:
Duplicate of #3951. Closing the old one as this has better explanation

PR #4133, opened 2024-01-23
Fixes problems with upload conflict policy (master)

xiroV:
Fixes the bugs described in #4120 .

Also improves the uploader by keeping file-descriptors open for some time (currently 10 seconds), before closing them.


provider-integration/integration-module/src/main/kotlin/plugins/storage/ucloud/NativeFS.kt
```diff
@@ -485,6 +497,9 @@ class NativeFS(
         val sourceParent = openFile(source.parent())
         val destinationParent = openFile(destination.parent())
 
+        log.debug("Opened ${source.path}")
+        log.debug("Opened ${destination.path}")
```
Tiggles, line 501: I think these could have more descriptive text, rather than both writing "Opened"

DanThrane (approved):

PR #4184, opened 2024-05-07
go code for gpfs

mrlhansen:


provider-integration/im2/pkg/im/gpfs/client.go
```diff
@@ -0,0 +1,324 @@
...
+		return false
+	}
+
+	// Wait
+	_, ok = c.JobWait(jr.Jobs[0].JobID)
+	return ok
+}
+
+func (c *Client) FilesetDelete(filesystem, fileset string) bool {
+	// Check variables
```
DanThrane, line 304:
```suggestion
	// Check variables
	// TODO(Dan): Check that filesystem and fileset does not contain '..' or '/'
```

mrlhansen:
Yes, it definitely needs some improvement, I just wanted to put it here for now.

PR #4239, opened 2024-05-30
Fixes #4204

### LINKED ISSUE #4204: Grant application page does not reload if id is changed (e.g. by clicking a notification while on the page)

Tiggles:
The Editor.tsx component is massive, and I'm not sure I have found all relevant points in the code, so I would like some extra eyes on this PR.

What it does:
- Makes a useEffect listen for `location.search` changes.
- Adds "Reset" event, to reset state for changing grants, otherwise GrantGivers seem to be duplicated if present in both
- Dispatch "Reset" event on `location.search` changes to ensure no state remains from the previously viewed grant


DanThrane (approved):
Code looks reasonable to me. This is how I believe it should work.

PR #4301, opened 2024-07-23
Grant Transfer update

### LINKED ISSUE #4279: Grant transfer does not work

hschu12:
Fixed #4279 . Also made comment section auto scroll down to newest revision or comment.

Only shows transfer options based on grant seekers available grant givers (not included those already in the grant).
Might be a better way for the frontend to show the possible transfer options than the new element "possibleTransfers" identical to "allcoators", but the "allocators" element was filtered and modified to only contain the allocators that was requested for resources in the grant, so it did not contain all possible grant givers of the grant seeker.

Not sure if it should be possible to transfer to a grant giver that is already participating in the grant. If it should be possible should the resources that overlap then just be summed or should something else happen?

### DIFF
--- a/frontend-web/webclient/app/Grants/Editor.tsx
+++ b/frontend-web/webclient/app/Grants/Editor.tsx
@@ -2189,7 +2189,7 @@ const CommentSection: React.FunctionComponent<{
     useLayoutEffect(() => {
         const scrollingBox = scrollingRef.current;
         if (!scrollingBox) return;
-        scrollingBox.scrollTop = scrollingBox.scrollHeight
+        scrollingBox.scrollTop = scrollingBox.scrollHeight;
     }, [entries]);
 
     const onKeyDown = useCallback<React.KeyboardEventHandler>(ev => {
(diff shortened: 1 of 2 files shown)


frontend-web/webclient/app/Grants/Editor.tsx
```diff
@@ -2331,7 +2342,7 @@ const GrantGiver: React.FunctionComponent<{
         if (!props.onTransfer) return;
 
         const result = await transferProject(
-            props.allAllocators.filter(it => !it.checked && it.id !== props.projectId)
+            props.transfers.filter(it => !it.checked && it.id !== props.projectId)
```
Tiggles, line 2345: Am I missing something or are `transfers` and `allAllocations` equal to each other?

DanThrane: They look identical to me. I am not sure if that is intended.

DanThrane (changes requested):
Some stylistic changes. Can you clarify your intend with `transfers` and `allAllocations` in the frontend?

frontend-web/webclient/app/Grants/Editor.tsx
Tiggles: As far as I can see, these can diverge when calling `loadRevision`, so I don't think they are guaranteed to be equal.

Tiggles: This is part of the initial comment, so maybe I should have read that first.

DanThrane (approved):

PR #4303, opened 2024-07-24
Fixes #4302

### LINKED ISSUE #4302: Re-position file-properties info for Folders
How it is now
image

Tiggles:
New look. While better, I think more should be done at some point. Maybe when doing more with metadata?
Screenshot 2024-07-24 at 14-48-19 UCloud _foasdoipa_foo-folder12


DanThrane (approved):
I agree, let's try to improve it further when we get more metadata features. Until then, I am happy with the new look.

Issue #4274, opened 2024-06-18
Improve storage scan after empty trash


hschu12:
When emptying trash we currently only scan the drive of the Trash. In project cases this is only the member files for the user that empties trash and in personal workspaces this is only Home. If files are deleted from other drives the changes are not applied.

Should check all drives for that given workspace.

mrlhansen:
Long term we should move to a solution where we have one Trash per drive instead.

PR #4311, opened 2024-08-01
Trashdevelopment

hschu12:
#4274 
Trashfolders are now created in the drive instead  in the member files. When emptying a trash folder every drive in the project is queue for scan.


Tiggles (approved):
Frontend seems fine.

frontend-web/webclient/app/UCloud/FilesApi.tsx
```diff
@@ -861,6 +888,13 @@ async function synchronizationOpOnClick(files: UFile[], cb: ResourceBrowseCallba
     snackbarStore.addSuccess(`${allSynchronized ? "Removed from" : "Added to"} Syncthing`, false);
 }
 
+export function isReadonly(entries: Permission[]): boolean {
+    const isAdmin = entries.includes("ADMIN");
+    const isEdit = entries.includes("EDIT");
```
DanThrane, line 893:
```suggestion
    const isAdmin = entries.includes("ADMIN");
    const isEdit = entries.includes("EDIT") || isAdmin;
```

frontend-web/webclient/app/UCloud/FilesApi.tsx
```diff
@@ -617,15 +630,21 @@ class FilesApi extends ResourceApi<UFile, ProductStorage, UFileSpecification,
...
+                    const isTrashDirectory = cb.directory?.status.icon == "DIRECTORY_TRASH"
                     if (!support) return false;
                     if (!(support as FileCollectionSupport).files.trashSupported) return false;
                     if ((support as FileCollectionSupport).files.isReadOnly) {
                         return "File system is read-only";
                     }
+                    if (
+                        (isTrashDirectory && cb.directory && isReadonly(cb.directory.permissions.myself)) ||
+                        (selected.length !== 0 && selected.every(it => it.status.icon === "DIRECTORY_TRASH" && isReadonly(it.permissions.myself)))) {
+                        return "You cannot delete readonly files."
```
DanThrane, line 642:
```suggestion
                        return "You cannot delete read-only files."
```

frontend-web/webclient/app/UCloud/FilesApi.tsx
```diff
@@ -597,7 +609,8 @@ class FilesApi extends ResourceApi<UFile, ProductStorage, UFileSpecification,
                 icon: "sensitivity",
                 enabled(selected, cb) {
                     const support = cb.collection?.status.resolvedSupport?.support;
-                    if ((support as FileCollectionSupport)) { /* ??? */}
+                    if ((support as FileCollectionSupport)) { /* ??? */
+                    }
```
DanThrane, line 613: I think this is just supposed to be a type-assertion. This will probably work (please check if it compiles):
```suggestion
                    support as FileCollectionSupport;
```

frontend-web/webclient/app/UCloud/FilesApi.tsx
```diff
@@ -636,10 +655,15 @@ class FilesApi extends ResourceApi<UFile, ProductStorage, UFileSpecification,
...
                         onConfirm: async () => {
-                            await cb.invokeCommand(
-                                this.emptyTrash(bulkRequestOf({id: cb.directory?.id ?? ""}))
-                            );
-                            cb.navigate("/drives");
+                            try {
+                                await cb.invokeCommand(
+                                    this.emptyTrash(bulkRequestOf({id: cb.directory?.id ?? ""}))
+                                );
+                                cb.navigate("/drives");
```
DanThrane, line 662: Would it be possible to change this such that it simply sends you back to the root of the drive instead of the drive view? I always found it a bit weird that we go all the way back to the drive listing.

frontend-web/webclient/app/UCloud/FilesApi.tsx
```diff
@@ -987,9 +1025,10 @@ export async function addFileSensitivityDialog(file: UFile, invokeCommand: Invok
     if (!isSensitivitySupported(file)) {
         dialogStore.addDialog(
             <>
-                <Heading.h2>Sensitive files not supported <Icon name="warning" color="errorMain" size="32" /></Heading.h2>
+                <Heading.h2>Sensitive files not supported <Icon name="warning" color="errorMain"
+                                                                size="32"/></Heading.h2>
```
DanThrane, line 1029: This is a nitpicking comment, sorry about that.

```suggestion
                <Heading.h2>
                    Sensitive files not supported <Icon name="warning" color="errorMain" size="32"/>
                </Heading.h2>
```

provider-integration/integration-module/src/main/kotlin/plugins/storage/ucloud/tasks/EmptyTrashTask.kt
```diff
@@ -89,7 +90,17 @@ class EmptyTrashTask(
...
             doWork = doWork@{ nextItem ->
                 val internalFile = pathConverter.ucloudToInternal(UCloudFile.create(nextItem.path))
+
                 val driveInfo = pathConverter.locator.resolveDriveByInternalFile(internalFile)
+
+                val owner = if (driveInfo.ownerIsUser == true) {
+                    WalletOwner.User(driveInfo.owner!!)
+                } else {
+                    WalletOwner.Project(driveInfo.owner!!)
+                }
```
DanThrane, line 100: This will crash on the vast majority of drives. There is an, unfortunately undocumented, reason that the `driveInfo.owner` is nullable. This is directly referencing a column in the database which is somewhat multipurpose and we cannot easily change it. The API here is admittedly quite bad. In any case, I wouldn't recommend inserting a not-null assertions unless you know that these things are not null.

The correct way to determine the workspace owner is by switching on the `UCloudDrive`. Is by passing the driveId to `DriveLocator.fetchMetadataForDrive` and using the `workspace` property it returns.

The owner information, in this specific property, is only valid for the following cases:

- The drive is owned by a personal workspace
- The drive is a legacy drive from projects (i.e. using a path with the format of `/projects/$Id/$RepositoryName`
- The drive is a member's file drive

DanThrane (changes requested):

frontend-web/webclient/app/UCloud/FilesApi.tsx
Tiggles: I was under the impression that we had discussed this a while ago and decided not to do it.

DanThrane: In that case we might have had a reason. Do you remember why we decided not to?

frontend-web/webclient/app/UCloud/FilesApi.tsx
Tiggles: Just as a note, this function has been moved, not changed, so this is how it was before this PR.

DanThrane: I noticed that this isn't part of the actual PR. I am pretty sure it should be changed to this though.

frontend-web/webclient/app/UCloud/FilesApi.tsx
hschu12: Jonas stated yesterday that it might be due to us not knowing how or where a provider does trash so by going to drives it was a safe bet.

DanThrane: In that case, we can definitely change to go to the root of the drive.

Tiggles: Something like this should work, I think. But maybe done prettier.
`cb.navigate(AppRoutes.files.path(pathComponents(cb?.directory)[0]));`

DanThrane (dismissed):

DanThrane (approved):

PR #4316, opened 2024-08-01
Implementation of missing file operations for slurm in go-im

### LINKED ISSUE #4280: Missing file operations in go-im
- [x] Upload of files
 - [x] Upload of folders
 - [x] Move to trash
 - [x] Empty trash
 - [x] Copy of files
 - [x] Copy of folders
 - [x] Task system
 - [x] bug: Move of folders does not respect rename policy

xiroV:
Implements the following missing file operations for slurm for go-im.

 - Upload of files over websockets
 - Upload of folders over websockets
 - Move to trash
 - Empty trash
 - Copy of files
 - Copy of folders
 - Task system

Fixes some minor bugs, including:

 - Move of folders does not respect rename policy
 - User not validated for upload/download sessions

fixes #4280


DanThrane (approved):
I had a quick look through the PR and I found no big issues. Given that this is pre-production code, I will just approve it and have a further look on the master branch myself. I think I want to do a bit of refactoring on some of the code. 

In particular, I think a few more files might be useful. I also think we might need to move some types, and come up with better guidelines for where certain types belong. I also think some of the types are too big (in bytes) and might need to pass parts of their structs by pointer to avoid copying huge structs arounds.

PR #4361, opened 2024-08-08
Fixes to application studio

### LINKED ISSUE #4305: Improvements to application studio
From a discussion with Frederica, I think there's a number of improvements we could make for the application studio to make it simpler. There also seem to be a few bugs.

 - [x] Deletion of application groups does not seem to work.
 - [x] Attempting to create an application with a title that already exist triggers an unhandled error, but even if we don't allow this, it is still possible (via. rename) to make multiple groups with the same title.
 - [x] The logo cache does not seem to update properly.
 - [x] The "Save changes" button does not give any feedback, but is also misleading since only _some_ of the parameters are saved when clicked (e.g. when an application are added to the group, it is saved directly).
 - [x] The drop-down to change the group of an application (from the app view) does not work. It was suggested that we simply remove the drop-down, since it makes more sense to just add applications to groups from the group view. It would be fine to still display the group in the app-view though.
 - [x] Delete buttons should use the hold-to-confirm button, instead of deleting the app/group without any confirmation.
 - [x] The mixed filter/create field on the groups list is confusing. It looks like a search bar, but pressing enter will create a new group. It would probably be better with separate fields for creation and searching.
 - [x] The list of groups are difficult to navigate. They seem to be ordered by creation time, but it would probably be better to sort by title.
 - [x] Project group permissions for applications does not seem to work at all, and it even seems like attempting to add this type of permission can make the ACL inaccessible.
 - [x] Is Public is not enforced?
 - [x] Deleting permissions from applications page trigger old type of modal confirmation, instead of hold-to-confirm.
 - [x] Confirm user exists before adding to ACL

xiroV:
fixes #4305


frontend-web/webclient/app/Applications/Studio/Applications.tsx
```diff
@@ -172,15 +107,29 @@ export const App: React.FunctionComponent = () => {
...
         fetchPermissionEntries(AppStore.retrieveAcl({name: name}));
-        setGroups(AppStore.browseGroups({itemsPerPage: 250}));
     }, [name]);
 
+    useEffect(() => {
+        setGroups(AppStore.browseGroups({itemsPerPage: 1000}));
+    }, [name]);
+
+    useEffect(() => {
+        console.log(selectedGroup);
```
Tiggles, line 117:
```suggestion
```

frontend-web/webclient/app/Applications/Studio/Applications.tsx
```diff
@@ -172,15 +107,29 @@ export const App: React.FunctionComponent = () => {
...
+        if (!apps.data.items[0]) return;
+        if (!selectedGroup) return;
+        if (selectedGroup === apps.data.items[0].metadata.group?.metadata.id) return;
+
+        invokeCommand(AppStore.assignApplicationToGroup({
+            group: selectedGroup,
+            name: apps.data.items[0].metadata.name
+        }));
+    }, [selectedGroup]);
```
Tiggles, line 126:
```suggestion
    useEffect(() => {
        const [firstApp] = apps.data.items;
        if (!firstApp) return;
        if (!selectedGroup) return;
        if (selectedGroup === firstApp.metadata.group?.metadata.id) return;

        invokeCommand(AppStore.assignApplicationToGroup({
            group: selectedGroup,
            name: firstApp.metadata.name
        }));
    }, [selectedGroup]);
```

frontend-web/webclient/app/Applications/Studio/Applications.tsx
```diff
@@ -212,18 +161,27 @@ export const App: React.FunctionComponent = () => {
     const projectEntityField = React.useRef<HTMLInputElement>(null);
     const groupEntityField = React.useRef<HTMLInputElement>(null);
 
+    const groupOptions = React.useMemo(() => {
+        return allGroups.map((appGroup) => ({
+            text: appGroup.specification.title,
+            value: appGroup.metadata.id
+        })).sort((a, b) => a.text > b.text ? 1 : -1)
+    }, [allGroups]);
+    console.log("groupOptions " + groupOptions.length);
```
Tiggles, line 170:
```suggestion
```

frontend-web/webclient/app/Applications/Studio/Applications.tsx
```diff
@@ -212,18 +161,27 @@ export const App: React.FunctionComponent = () => {
     const projectEntityField = React.useRef<HTMLInputElement>(null);
     const groupEntityField = React.useRef<HTMLInputElement>(null);
 
+    const groupOptions = React.useMemo(() => {
+        return allGroups.map((appGroup) => ({
+            text: appGroup.specification.title,
+            value: appGroup.metadata.id
+        })).sort((a, b) => a.text > b.text ? 1 : -1)
```
Tiggles, line 168: I don't _really_ understand comparisons for arrays, but this should be similar.

```suggestion
        })).sort((a, b) => a.localeCompare(b))
```

frontend-web/webclient/app/Applications/Studio/Applications.tsx
```diff
@@ -244,17 +201,23 @@ export const App: React.FunctionComponent = () => {
...
+                            <Label>Group</Label>
+                            <Box mb="20px" textAlign="right">
+                                    <Select
+                                        name="group" 
+                                        value={selectedGroup}
+                                        onChange={(event) => setSelectedGroup(event.target.value)}
+                                    >
+                                        {groupOptions.map(it => (
+                                            <option value={it.value}>{it.text}</option>
+                                        ))}
```
Tiggles, line 213: Small thing, this selects the first option by default, right? Would it make sense to have an empty options, so UCloud Admins don't select the first group by mistake?

xiroV: Yes, I actually thought about it, but forgot to add it :+1:

frontend-web/webclient/app/Applications/Studio/Applications.tsx
```diff
@@ -415,25 +378,17 @@ export const App: React.FunctionComponent = () => {
...
-                                                                    )}
-                                                                    </Text>
-                                                                </Box>
-                                                            ),
-                                                            onConfirm: async () => {
+                                                <TableCell>
+                                                    <Flex justifyContent="right">
+                                                        <ConfirmationButton
+                                                            icon="heroTrash"
+                                                            actionKey={
```
Tiggles, line 385: I don't think this is needed as `onAction` uses no arguments.

frontend-web/webclient/app/ui-components/Input.tsx
```diff
@@ -160,17 +160,13 @@ const InputLabelClass = injectStyle("input-label", k => `
     }
     
     ${k}[data-left-label="true"] {
-        border-top-left-radius: 5px;
-        border-bottom-left-radius: 5px;
```
Tiggles, line 164: These are used in `CreateProduct.tsx` and `DataList.tsx`. Is this a mistake?

frontend-web/webclient/app/Applications/Studio/Group.tsx
```diff
@@ -215,6 +191,40 @@ export const AppGroup: React.FunctionComponent = () => {
...
+                                    <Button type="button" onClick={async () => {
+                                        if (!group.data) return;
+
+                                        const titleField = groupTitleField.current;
+                                        if (titleField === null) return;
+
+                                        const newTitle = titleField.value;
+                                        if (newTitle === "") {
+                                            snackbarStore.addFailure("Title cannot be empty", false);
+                                            return
```
Tiggles, line 204:
```suggestion
                                            return;
```

frontend-web/webclient/app/Applications/Studio/Group.tsx
```diff
@@ -215,6 +191,40 @@ export const AppGroup: React.FunctionComponent = () => {
...
+                                        }
+
+                                        const descriptionField = groupDescriptionField.current;
+                                        if (descriptionField === null) return;
+
+                                        const newDescription = descriptionField.value;
+
+                                        const success = await invokeCommand(AppStore.updateGroup({
+                                            id: group.data?.metadata.id,
```
Tiggles, line 213: Do we want to submit if `id` is `null/undefined`?

frontend-web/webclient/app/Applications/Studio/Group.tsx
```diff
@@ -215,6 +191,40 @@ export const AppGroup: React.FunctionComponent = () => {
...
+                                            newTitle: newTitle,
+                                            newDescription: newDescription,
+                                            newDefaultFlavor: defaultApplication,
+                                            newLogoHasText: logoHasText
+                                            // tags
```
Tiggles, line 218: Something that needs to be done?

frontend-web/webclient/app/Applications/Studio/Group.tsx
```diff
@@ -142,6 +142,7 @@ export const AppGroup: React.FunctionComponent = () => {
                             {uniqueApps.map(app => (
                                 group.data?.status?.applications?.map(app => app.metadata.name).includes(app) ? null : (
                                     <ListRow
+                                        key={app}
```
Tiggles, line 145: Your choice, but I would probably use `appName` instead.

frontend-web/webclient/app/Applications/Studio/Group.tsx
xiroV: If you mean the `// tags` comment, it was already there. Probably from the old app store, but not sure.

Tiggles: Okay, then I'll mark this as resolved.

frontend-web/webclient/app/ui-components/Input.tsx
xiroV: I don't see any places where both rightLabel and leftLabel are being used, except for the permissions-editor for applications, but either way I don't think the previous styling was correct. The border radius is already 5, and also this prevented adding both rightLabel and leftLabel at the same time, since they would overwrite each other's border-radius.

Tiggles: I think you're correct. Marking as resolved.

Tiggles (approved):

DanThrane (approved):

PR #4369, opened 2024-08-13
Upgrade to deicReport

hschu12:
Upgraded extract script to be able to run with new DB and from single command


hschu12:
Not perfect, but way better. Single command to create all reports.

Issue #4360, opened 2024-08-08
Store file "clipboard" outside of the component


DanThrane:
This should allow for copying & pasting files across component reloads. It does not have to be persistent across page reloads, but could be (maybe add a max age in that case).

Tiggles:
Some things to maybe consider:

- Should the copy paste also allow to happen between modals and non-modal components (and vice versa)?
- When should it be removed from the clipboard? E.g. leaving a specific page, accessing specific pages, etc. (age already mentioned)

DanThrane:
> Should the copy paste also allow to happen between modals and non-modal components (and vice versa)?

I don't see any reason why it shouldn't be allowed.

> When should it be removed from the clipboard? E.g. leaving a specific page, accessing specific pages, etc. (age already mentioned)

For a normal copy, I don't think it should unless overriden by a new copy. For cut it should disappear when pasted. We could consider a maximum life-time of a few minutes, but beyond that I don't think it should disappear for any reason related to the UI.

Tiggles:
Good. I agree with both of these.

Tiggles:
Is this considered finished with the feature flag in place? @DanThrane

PR #4373, opened 2024-08-14
Attempt at #4360

Tiggles:
So, this attempt at #4360 does not store it outside the component, but stores it as a static variable inside the ResourceBrowser component, namespaced by the parent-component, e.g. "File", "Drive", etc.

Currently, it resets on page reload, so clipboard actions can be stored forever, as long as they are not invoked, or the page is reloaded.

This is a draft, as I think some discussion based on the solution should probably be discussed, and I have probably missed some cases that will invoke the undo/redo in the wrong way.


Tiggles:
Just tried:

- Copy a file in FileBrowse (ctrl/cmd+C)
- Go to App stores (unmounting FileBrowse)
- Go to FileBrowse, click "Move to..." or "Copy to..."
- Enter folder
- Paste (ctrl/cmd+V)

Works!

DanThrane (approved):

PR #4376, opened 2024-08-16
[IM2] Jline2/prompt functionality

### LINKED ISSUE #4320: [IM2/Systems] jline2 alternative in Go
Create an alternative implementation of jline2 in Go. This will be required by various CLI tools (such as the `ucloud slurm config bootstrap` command). This issue can be solved either by finding a similar library or completely re-implementing it. Both seem perfectly feasible since this functionality isn't _that_ complicated.

xiroV:
Contains the following functionality for building simple CLI prompts:

 - Prompt for text.
 - Prompt for confirmation (yes/no).
 - Menu, which can work as both a single-select prompt or multi-select prompt.
 - Loading/spinning indicator.

Examples in `main.go`.

fixes #4320


DanThrane (approved):
Looks good to me, appears to have everything we need 😀

PR #4379, opened 2024-08-22
Fixed issue with PI being grant giver and not reciever on gift.

hschu12:


DanThrane (changes requested):
This seems like a security hole waiting to happen. I think we need to find a way to fix this issue that doesn't involve changing the permission model.

It seems way too likely that we will later do a `SystemOnBehalfOfUser` expecting it to check the permissions, but now suddenly it will always pass through the `isGrantGiver()` check.

DanThrane (approved):

PR #4435, opened 2024-10-02
gdu instead of du

hschu12:
Added element to config that allows to use gdu instaed of du to do storage scan of drives if installed.


mrlhansen:
I think you need `-a` for `gdu` otherwise it will not report the same number as `du`.

PR #4485, opened 2024-11-05
fixes color issue

### LINKED ISSUE #4451: Wrong coloring for allocations
Not sure if it's a front-end or back-end bug
Screenshot 2024-10-24 at 09 08 53

hschu12:
Fixes #4451 

Not posssible to show limitation of resource on the overview of a sub-project since the info is not available.
See Note from Dan on line 1011 in changed file (index.ts)

Screenshot 2024-11-05 at 13 30 47


Tiggles:
We should still do _something_ wrt. the aggregation row, as an empty one with wrong information might be worse than not having it. A different issue, though.

Issue #4469, opened 2024-10-31
[IM2] Dynamic registration of web/VNC interfaces


DanThrane:
This issue tracks the implementation of dynamic registration of web/VNC interfaces. This will allow service providers
to dynamically add interactive interfaces after the application has started. Use cases include:

- Using the new feature for batch applications, to create custom web/vnc applications which UCloud do not know of ahead of time
- Starting a custom VNC/web server in generic interactive applications such as starting your own web-application in coder/terminal
- Alternative solution to #3709

## Strategy

- Expand the "openInteractiveSession" call to take a nullable "target" string. By default, this will be null and use the current behavior. If it is not null, then this will use a different "target" which is forwarded to the service provider
- The service provider should be expanded to use these targets
- Notify the frontend by sending a textual update in a specific format. This format should notify about the name and rank   for the interface. This should use the same strategy as we use for SSH. No new calls are required for this and only needs changes to the frontend.

This issue does not require integration in the integration module yet. But one could imagine extending the "workflow" feature with a new method which actually triggers the registration. For example, adding the following to a workflow might automatically cause an interface registration:

```
# dynamicInterface(rank, targetName, targetPort)
{{ dynamicInterface(0, "web", 8080) }}
```

Issue #874, opened 2019-07-03
Provide info on which state failed for job


Tiggles:
Instead of showing every step failed, show specifically which one failed.

PR #4523, opened 2024-12-16
Frontend for dynamic registration of VNC/web interfaces 

xiroV:
Part of #4469


frontend-web/webclient/app/UCloud/JobsApi.tsx
```diff
@@ -154,6 +154,8 @@ export interface OpenInteractiveSessionRequest {
     id: string;
     rank: number;
     sessionType: "WEB" | "VNC" | "SHELL";
+    target?: string | null;
```
Tiggles, line 157: Can this really be undefined, string or null, or just two of the three?

frontend-web/webclient/app/ui-components/Button.tsx
```diff
@@ -102,6 +103,11 @@ export const ButtonClass = injectStyle("button", k => `
         border-bottom-left-radius: 0;
     }
 
+    ${k}[data-attached-left=true] {
+        border-top-right-radius: 0;
+        border-bottom-right-radius: 0;
+    }
```
Tiggles, line 109: Does this mean this is attached to the left of the button?

Also, if `attached-left` exists, attached should be renamed to `attached-right`

frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -723,22 +743,37 @@ function isSupported(jobBackend: string | undefined, support: ComputeSupport | u
...
-                    {" "}
-                    <Box style={{display: "inline"}} color={"textSecondary"}>(ID: {job.id})</Box>
-                </Heading.h2>
-                <Box flexGrow={1} />
-                <div><CancelButton job={job} state={"RUNNING"} /></div>
-            </Flex>
-            <RunningButtonGroup job={job} rank={0} redirectToWeb={interfaceLinks[0]} />
-        </Flex>
-    </>;
-};
```
Tiggles, line 741: Why has this been removed instead of being in a separate component?

frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -827,11 +862,104 @@ function parseUpdatesForAccess(updates: JobUpdate[]): ParsedSshAccess | null {
     return null;
 }
 
+interface InterfaceTarget {
+    rank: number,
+    type: "WEB" | "VNC" | "SHELL";
```
Tiggles, line 867: Is there any cases with `SHELL` that will be added to the results?

frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -1324,49 +1451,11 @@ function getAppType(job: Job): string {
...
-
-            (async () => {
-                try {
-                    const result = await call<BulkResponse<InteractiveSession>>(JobsApi.openInteractiveSession(bulkRequestOf(...requests)));
-                    setLinks(links => {
-                        for (const res of result?.responses ?? []) {
-                            links[res.session.rank] = (res.session as WebSession).redirectClientTo
-                        }
-                        return links;
-                    })
```
Tiggles, line 1354: wrt. what we talked about yesterday @xiroV, `setLinks` from the `useState` above is being used here after `results` are fetched. This is why this works, as it forces a re-render.

frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -827,11 +862,104 @@ function parseUpdatesForAccess(updates: JobUpdate[]): ParsedSshAccess | null {
...
+                });
+            }
+        } catch (e) {
+            console.warn(e);
+        }
+    }
+
+    return result;
+}
+
```
Tiggles, line 957: This function can be improved by not awaiting immediately after calling the backend, so both promises can be awaited on at the same time. Not an urgent fix, but a `TODO` could be added.

frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -1399,37 +1487,48 @@ const RunningButtonGroup: React.FunctionComponent<{
...
+                    trigger={
+                        <div className={InterfaceSelectorTrigger}> 
+                            <Icon name="chevronDownLight"/>
+                        </div>
+                    }
+                    right={"0px"}
+                    minWidth={"245px"}
+                    paddingControlledByContent
+                >
+                    {interfaceLinks.slice(1).map((link, k) => (
```
Tiggles, line 1508: I would filter out empty links. A link to a non-existant location seems like a weird user experience.

`interfaceLinks.slice(1).filter(it => it.link).map((link, k) => (`

frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -1399,37 +1487,48 @@ const RunningButtonGroup: React.FunctionComponent<{
...
+                                key={link.link ?? (link.rank + (link.port?.toString() ?? ""))}
+                            >
+                                <Flex
+                                    key={k}
+                                    gap="5px"
+                                    alignItems={"center"}
+                                    p={8}
+                                >
+                                    <Icon name="heroArrowTopRightOnSquare" />
+                                    {link.target ?? "Open interface"}
```
Tiggles, line 1521: Depending on the length of the `link.target` value, it might make sense to wrap it in a `Truncate`-tag.

Tiggles (changes requested):
Some small changes requested

frontend-web/webclient/app/UCloud/JobsApi.tsx
DanThrane: It can be all three.

frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -827,11 +860,103 @@ function parseUpdatesForAccess(updates: JobUpdate[]): ParsedSshAccess | null {
...
+                const supportsInterface =
+                    (parsedTarget.type === "WEB" && isSupported(backendType, support, "web")) ||
+                    (parsedTarget.type === "VNC" && isSupported(backendType, support, "vnc"));
+
+                if (parsedTarget.type === "WEB" && supportsInterface && job.status.state === "RUNNING") {
+                    const requests: OpenInteractiveSessionRequest[] = Array(job.specification.replicas)
+                        .fill(0).map((_, i) => ({sessionType: "WEB", id: job.id, rank: i, target: parsedTarget.target, port: parsedTarget.port}));
+
+                    try {
+                        const sessionResult = await invokeCommand<BulkResponse<InteractiveSession>>(JobsApi.openInteractiveSession(bulkRequestOf(...requests)));
```
DanThrane, line 900: Collect multiple requests in an array and then do a single call after the outer loop finishes. Include even the normal calls in this request, as this will also resolve the comment from Jonas.

frontend-web/webclient/app/Applications/Jobs/View.tsx
DanThrane: The job of `findInterfaceTargets` is to populate the `link` property and should never be undefined at this point.

frontend-web/webclient/app/Applications/Jobs/View.tsx
DanThrane: Jonas has a valid point given that `link.target` is coming from a user-specified value. It might be very long.

frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -319,9 +320,15 @@ export function View(props: {id?: string; embedded?: boolean;}): React.ReactNode
...
     const [jobFetcher, fetchJob] = useCloudAPI<Job | undefined>({noop: true}, undefined);
+    const [, invokeCommand] = useCloudCommand();
     const job = jobFetcher.data;
+    const [interfaceTargets, setInterfaceTargets] = useState<InterfaceTarget[]>([]);
 
-    const interfaceLinks = useRedirectToWebLinks(job);
+    useEffect(() => {
+        if (!job) return;
+
+        findInterfaceTargets(job, invokeCommand).then(setInterfaceTargets);
```
DanThrane, line 330: I haven't checked, but it wouldn't surprise me if this function is never reinvoked when new updates arrive over the websocket connection. This function might need to live closer to the other update parsing code.

frontend-web/webclient/app/Applications/Jobs/View.tsx
DanThrane: The IM2 implementation will never produce a value of `"SHELL"` so it is probably safe to remove from the list here.

frontend-web/webclient/app/Applications/Jobs/View.tsx
xiroV: I don't remember exactly, but it might have been an attempt to simplify/flatten the hierarchy a bit. I'll revert it back though.

frontend-web/webclient/app/Applications/Jobs/View.tsx
Tiggles: Then line #874 should be changed accordingly.

frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -1378,58 +1548,75 @@ const RunningButtonGroup: React.FunctionComponent<{
         (appType === "WEB" && isSupported(backendType, support, "web")) ||
         (appType === "VNC" && isSupported(backendType, support, "vnc"));
 
+    const searchableInterfaceLinks: SearchableInterfaceTarget[] = interfaceLinks.map(it => ({
+        searchString: it.target + " " + (it.rank+1), ...it
+    }));
+
+    const searchableTerminalLinks: SearchableTerminalTarget[] = terminalLinks.map(it => ({
```
Tiggles, line 1555: [Non-blocking] Formatter needs to be run

frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -1378,58 +1548,75 @@ const RunningButtonGroup: React.FunctionComponent<{
...
+        searchString: (it.rank+1).toString(), ...it
```
Tiggles, line 1556: [Non-blocking] Formatter needs to be run

frontend-web/webclient/app/ui-components/RichSelect.tsx
```diff
@@ -48,11 +49,15 @@ export function RichSelect<T, K extends keyof T>(props: {
...
-            </div>
+        trigger={props.FullRenderSelected ? 
+                <props.FullRenderSelected element={props.selected} onSelect={doNothing} />
+            :
+                props.RenderSelected ? 
+                    <div className={TriggerClass} ref={triggerRef}>
+                        <props.RenderSelected element={props.selected} onSelect={doNothing}/>
+                        <Icon name="chevronDownLight"/>
+                    </div>
+                : <></>
```
Tiggles, line 60: [Non-blocking] If I'm reading this correctly, only one of `FullRenderSelected` and `RenderSelected` will be rendered. Can't you just pass the full component down once instead of allowing to different?

frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -2,20 +2,20 @@ import * as React from "react";
...
-import {Box, Button, Card, ExternalLink, Flex, Icon, Link, Truncate} from "@/ui-components";
+import {Box, Button, Card, ExternalLink, Flex, Icon, Link, Select, Truncate} from "@/ui-components";
 import TitledCard from "@/ui-components/HighlightedCard";
 import {buildQueryString, getQueryParamOrElse} from "@/Utilities/URIUtilities";
 import {device, deviceBreakpoint} from "@/ui-components/Hide";
 import {CSSTransition} from "react-transition-group";
 import {Client, WSFactory} from "@/Authentication/HttpClientInstance";
 import {dateToString, dateToTimeOfDayString} from "@/Utilities/DateUtilities";
-import {MarginProps} from "styled-system";
+import {MarginProps, minWidth, right} from "styled-system";
```
Tiggles, line 18: I think these were imported by mistake

frontend-web/webclient/app/Applications/Jobs/View.tsx
```diff
@@ -1324,112 +1444,177 @@ function getAppType(job: Job): string {
...
-                <Button width="220px">
-                    <Icon name={"heroArrowTopRightOnSquare"} />
-                    <div>Open interface {hasMultipleNodes ? `(node ${rank + 1})` : null}</div>
-                </Button>
-            </Link>
+
+        {interfaceLinks.length < 1 ? <></> : (
+            <Flex>
+                <Link to={interfaceLinks[defaultInterfaceId]?.link ?? ""} aria-disabled={!interfaceLinks[defaultInterfaceId]} target={"_blank"}>
+                    <Button attachedLeft={interfaceLinks.length > 1} disabled={!interfaceLinks[0]}>
```
Tiggles, line 1597: Is index 0 the correct index to access here? `defaultInterfaceId` is used around it.

Tiggles (commented):
One question I need answered. Other than that, I believe it's fine.

Tiggles (approved):

Issue #4468, opened 2024-10-31
[IM2] Status page feature


DanThrane:
This issue tracks adding a status page feature to the new integration module. In the integration module, this should
expose a fairly simple static webpage which shows information about the system and some information about the local
user, such as their username. The core should be able to request a link from the service provider which will be shown
somewhere in the user-interface. The service provider is allowed to not implement this endpoint (i.e. returning 404),
in this case the core should assume they do not have one.

Jonas has already implemented a prototype for the page. Concrete details for what we put in the page should be somewhat
configurable, Martin and Jonas probably both have an opnion on what should be configurable.

PR #4568, opened 2025-01-24
Preperation for provider status/condition + minor config improvement

xiroV:
Part of #4468 

Adds status (called *condition*) retrieve end-points, and shows an icon next to the provider name on the dashboard if a status was retrievable (otherwise no icon will be displayed). Only the go-provider implements this end-point, and will currently always return that it's available, along with a link to https://status.cloud.sdu.dk. The mechanism for changing this from the provider-side, as well as integration with the status template, will be implemented later.

Additionally, a small change to the configuration reader, allows configuration files with either `.yml` or `.yaml` file extensions (`.yml` will be preferred). Only configuration files with the `.yml` extension were considered before.


frontend-web/webclient/app/Providers/Connect.tsx
```diff
@@ -17,18 +17,64 @@ import {connectionState} from "./ConnectionState";
...
+            tooltip = "Unavailable"
+            break;
+        case "MAINTENANCE":
+            icon = "heroWrench";
+            color = "textSecondary";
+            tooltip = "Down for maintenance"
+            break;
+    }
+
+    return <Link to={props.condition.page ?? ""} target={"_blank"}>
```
Tiggles, line 66: We shouldn't have empty hrefs/tos

frontend-web/webclient/app/Providers/Connect.tsx
```diff
@@ -39,6 +85,20 @@ export const Connect: React.FunctionComponent<{embedded?: boolean}> = props => {
...
+            callAPI(
+                apiRetrieve(
+                    {provider: provider.provider},
+                    "/api/providers/integration",
+                    "condition"
+                )
+            ).then(condition => {
+                setProviderConditions((prev) => new Map(prev.set(provider.providerTitle, condition)));
+            });
+        });
```
Tiggles, line 99: Shouldn't integration condition just accept a bulk request in the backend?

DanThrane (approved):
I agree with the comments from @Tiggles. But I think it is better than we do a larger revision later this year when we look at the refactoring we intend to do with the Core.

DanThrane (approved):

PR #4635, opened 2025-02-28
Epic 4548

Tiggles:
- [X] Add "Close saved tabs" operation"
- [X] Scroll to tab opening it, if not in view
- [X] Warn on close for scripts editor when having unsaved files
- [X] Actually save for vim-bindings "save"/"save-and-quit"
    - [X] Preview editor
    - [X] Scripts/workflow editor
- [X] Add VIM status-bar if switched on
- [X] Add jinja2 to the language selector dropdown
- [X] Header/toolbar is too wide for modal context
- [X] Passed language can get overriden (the scripts language passed is ignored)
- [X] Default to readonly, opt-in to allow editing - with "do ˚not show again" setting
- [X] Renaming an edited (but˚ not saved) file, will revert contents to initial version
- [X] Renaming and edited file does not update dirty files set correctly
- [X] Remove VIM support for now
- [X] Investigate VIM package from VSCode/VSCodium
- [X] Syntax selector is too small (see reStructuredText)
- [X] Syntax selector dropdown is on the left side, but should be able to be directly below
- [X] Language icons aren't resolved correctly on sandboxoierjgoerjg
- [X] Arrow-key navigation is missing from syntax selector
- [X] When leaving site, warn on dirty file not saved


DanThrane (approved):
Looks good to me.

PR #4637, opened 2025-03-04
Prometheus metrics

### LINKED ISSUE #4585: [IM2-K8s] Add Prometheus metrics to existing code
Add prometheus metrics to the existing code. Try to get all existing metrics and add new relevant ones to the code base.

An example of a metric in the old codebase: https://github.com/sdu-escience/ucloud/blob/c04385e0a79fa62ba22094967b6a3ae5ecd8a4c3/provider-integration/integration-module/src/main/kotlin/plugins/storage/ucloud/Plugin.kt#L1103

An example of a metric in the new codebase: https://github.com/sdu-escience/ucloud/blob/71093231f13570a9164fef6d7658b6c7ffe5347f/provider-integration/im2/pkg/im/controller/connection.go#L652

xiroV:
fixes #4585


provider-integration/im2/pkg/im/controller/job_database.go
```diff
@@ -100,11 +100,39 @@ func InitJobDatabase() {
...
 	initIpDatabase()
+
+	// Job metrics
+	go func() {
+		for util.IsAlive {
+			var jobsRunning float64 = 0
+			var jobsInQueue float64 = 0
+			var jobsSuspended float64 = 0
+
+			for _, job := range activeJobs {
```
DanThrane, line 111: Missing mutex, use the existing one..

provider-integration/im2/pkg/im/controller/manager.go
```diff
@@ -184,6 +239,7 @@ func sendResponseOrError(w http.ResponseWriter, data any, err error) {
 }
 
 func sendError(w http.ResponseWriter, err error) {
+	metricRequestErrorCounter.Inc()
```
DanThrane, line 242: I think we should differentiate a bit more between the errors. At least if it is 4xx or 5xx.

DanThrane (changes requested):

DanThrane (dismissed):

DanThrane (approved):

PR #4638, opened 2025-03-05
Legacy support for file sensitivity for im2-k8s

### LINKED ISSUE #4587: [IM2-K8s] Legacy sensitivity
We still have some files around which uses the legacy file sensitivity system. We should continue to support this by reading the extended attributes of files to determine what the legacy sensitivity is.

https://github.com/sdu-escience/ucloud/blob/6939f64075072eda923a35875039b5c92791a97d/provider-integration/integration-module/src/main/kotlin/plugins/storage/ucloud/FileQueries.kt#L100

xiroV:
fixes #4587

### DIFF
--- a/frontend-web/webclient/app/UCloud/FilesApi.tsx
+++ b/frontend-web/webclient/app/UCloud/FilesApi.tsx
@@ -108,6 +108,7 @@ export function isSensitivitySupported(resource: UFile): boolean {
         switch (resource.specification.product.provider) {
             case "k8":
             case "K8":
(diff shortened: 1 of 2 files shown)


DanThrane (commented):
I should have seen this the first time I looked at it. Can you insert a check which ensures that we are only sending back valid values? The property can be set by users and ideally we don't want to use them if they are invalid. The Core will most likely complain a lot in case it is set to an invalid value.

DanThrane (approved):

Issue #4792, opened 2025-05-06
File search doesn't leave any obvious way to return to normal browsing


xiroV:
After using file search, there doesn't appear to be any obvious way to return to "normal" file browsing, except for browsing to the path of the drive again.

One way could be searching for an empty string, which currently seems to return everything in the drive, however I think a button would make it more obvious.

Tiggles:
Just closing the search field should go back, I think

xiroV:
It doesn't go back.

But actually I thought that icon was the submit-button, not for closing the search field.

Tiggles:
Yeah, sorry, I meant clicking the search icon again (closing it) could be the behavior to go back, not that that is how it works.

Tiggles:
> Yeah, sorry, I meant clicking the search icon again (closing it) could be the behavior to go back, not that that is how it works.

Turns out, this IS the intended behavior, but we haven't implemented a "searchHidden" function for the derived ResourceBrowser components.

Tiggles:
Currently, clicking the "Drives" title-breadcrumb-thing will revert to showing initial files/drives

PR #4829, opened 2025-05-14
Issue #4792

Tiggles:


frontend-web/webclient/app/Files/FileBrowse.tsx
```diff
@@ -1382,6 +1382,10 @@ function FileBrowse({opts}: {opts?: ResourceBrowserOpts<UFile> & AdditionalResou
                     )
                 });
 
+                browser.on("searchHidden", () => {
+                    browser.open(lastActiveFilePath, true);
```
DanThrane, line 1386: I don't fully remember the semantics of `searchHidden`.  So I have two questions:

1. Can this trigger without searching (i.e. just clicking twice on the search icon)? Does this trigger the expected behavior?
2. If this page was opened on the search page, then it appears that `lastActiveFilePath` is `""`. What does `browser.open("", true)` do and does it make sense?

Tiggles: 1. Yes, double clicking will trigger this and use the empty string, so this PR should not be accepted in its current state.
2. What "search page" are you referring to?

DanThrane: Search and refresh on that page (with the new URL)

Tiggles: Seems to revert to `app/drives` in both cases for me.

DanThrane (approved):

PR #4835, opened 2025-05-14
Fixes #4766

### LINKED ISSUE #4766: [Shares] Accepting a username share, while having active project, redirects to dashboard with explanation

Tiggles:


DanThrane (commented):
I don't understand why this requirement is needed.

Tiggles:
You mean the requirement of being in 'My workspace' to view/accept shares?

DanThrane:
Yes, I don't understand why the frontend cannot accept a share in a project. I can partially understand not being able to view it, but accepting does not make sense to me.

Tiggles:
I've tried looking at the backend code, but I can't find any reference to it, but it seems to me that this project/my workspace share restriction is a backend thing. The same request, with different project-headers yield different responses.

DanThrane:
In that case, can we change it to just switch to your personal workspace when you are accepting a share? Especially when doing this through a link, since I don't think that will even feel wrong.

Tiggles:
I think we are talking about different things, I must have glossed over the reference to 'link' the first time.

This is when clicking a notification.

DanThrane:
In that case I would just do in generally. I noticed that this doesn't work when doing it through a link.

Tiggles:
Just to make sure:

- User A creates a link for Sharing and sends it to User B.
- User B has an active project and opens the link.

This does not work for you?

DanThrane (approved):

PR #4845, opened 2025-05-15
Fixes #4844

### LINKED ISSUE #4844: User-interface does not show other member files in "Add folder" drive dropdown
Steps to reproduce:

1. Open the drive list in a project as an admin/PI (which has other members)
2. _Without_ reloading the browser go to an application page an attempt to mount another member's "member drive"

In this case, the drives will not appear. However, if you reload on the application page, it is possible to select another member's "member drive". This appears to be caused by the wrong calls being made to the backend.

Tiggles:


Tiggles:
Setting `filterMemberFiles` to `DONT_FILTER_COLLECTIONS` will make `browse` set it to `SHOW_ONLY_MINE`.
Setting it to `all` will make `browse` set it to `DONT_FILTER_COLLECTIONS`.
```
    browse(
        req: PaginationRequestV2 & FileCollectionFlags & SortFlags
    ): APICallParameters<PaginationRequestV2 & FileCollectionFlags, PageV2<FileCollection>> {
        if (req["filterMemberFiles"] == "all") {
            req["filterMemberFiles"] = "DONT_FILTER_COLLECTIONS";
        } else if (req["filterMemberFiles"] === "true") {
            req["filterMemberFiles"] = "SHOW_ONLY_MEMBER_FILES";
        } else {
            req["filterMemberFiles"] = "SHOW_ONLY_MINE";
        }
        return super.browse(req);
    }```

DanThrane (approved):

PR #4943, opened 2025-08-18
Elastic audit Go code

hschu12:
New backend code for  #4720 including
- Deletion of  old indices and expired log entires
- Shrinking indices code
- Reindex indices code
- Grafana alias generation
- Misc utility code

Does not include old kotlin
- Backup code (was never used)
- Setup code (Handled by Martin now)
- Remove Flood code (only had the use for it once)

The code still need invokation code to be launched by k8  jobs or inline timer
For cleanup use CleanUpLogs() once per day
For grafana alias creation use CreateGrafanaAliases() every 3 hours

Most calls have been tested locally on single node cluster with few indices and low doc count.
Error might show up on production but errors are only logged and not throwing / panicing


DanThrane (changes requested):
Change the code to improve consistency across the project. This is mostly minor syntactical differences.

Avoid exporting functions when not needed. Keep in mind that these are in the entire package so add prefixes to types when needed.

All the log functions use the wrong format.

The code doesn't compile. Running tests near the package would cause to get angry at the incorrectly used log functions.

Don't ignore warnings from the IDE/compiler. There are many unhandled errors in pretty much all of the API calls to ElasticSearch. Consider an overall strategy and try to execute it as local as possible.

Consider adding batching and retries in case that ElasticSearch is unavailable.

JSON marshaling and unmarshaling is often done in an unnecessarily complicated way, I have added examples on how to improve it.

DanThrane (dismissed):

PR #4954, opened 2025-09-01
PR response

hschu12:
Edits to former PR

### DIFF
--- a/frontend-web/webclient/app/Accounting/Usage.tsx
+++ b/frontend-web/webclient/app/Accounting/Usage.tsx
@@ -344,19 +344,20 @@ function stateReducer(state: State, action: UIAction): State {
         }
 
         let jobUsageByUsers: UsagePerUser
-        state.remoteData.chartData?.usagePerUser
 
         if (state.remoteData.chartData?.usagePerUser === undefined) {
             jobUsageByUsers = emptyUsagePerUserList
         } else {
-            const usage = state.remoteData.chartData?.usagePerUser!!.data
+            const usage = state.remoteData.chartData?.usagePerUser!.data
             jobUsageByUsers = {dataPoints: usage}
         }
 
         const availableUnitsSet = unitsFromSummaries(state.summaries);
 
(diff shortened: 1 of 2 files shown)


frontend-web/webclient/app/Accounting/Usage.tsx
```diff
@@ -1413,7 +1414,7 @@ const UsageByUsers: React.FunctionComponent<{loading: boolean, data?: UsagePerUs
                     </thead>
                     <tbody>
                         {data.dataPoints.map(it =>
-                            <tr>
+                            <tr key={it.username + " - " + it.category.name + " (" + getShortProviderTitle(it.category.provider) + ")"}>
```
Tiggles, line 1417: For future reference, this is not readable for anybody other than React internally. Should happen with matching, so shorter is usually better.

For this case, it's probably fine as is.

frontend-web/webclient/app/Accounting/Usage.tsx
```diff
@@ -344,19 +344,20 @@ function stateReducer(state: State, action: UIAction): State {
...
-            const usage = state.remoteData.chartData?.usagePerUser!!.data
+            const usage = state.remoteData.chartData?.usagePerUser!.data
             jobUsageByUsers = {dataPoints: usage}
         }
 
         const availableUnitsSet = unitsFromSummaries(state.summaries);
 
-        availableUnitsSet.push(JOBS_UNIT_NAME);
-
+        //TODO() Need better if statement, but at the moment this is the only thing that is shown on this page
```
Tiggles, line 357: Are the parentheses empty, as you don't know who to add?

Tiggles (approved):

frontend-web/webclient/app/Accounting/Usage.tsx
Tiggles: Yes, you'd likely need just something like `{it.username + it.category.name}` to ensure uniqueness in this case.

DanThrane (approved):

PR #4972, opened 2025-09-17
Frontend: fixes #4971 and moves warning triangle above progress bar.

### LINKED ISSUE #4971: Negative 0 usage on resource overview and Dashboard
Both seen on the Dashboard and in the allocation overview page. Json seems fine with no negative usage, so I guess its a normalization on the frontend that somehow flips 0 to become -0. Frontend should perhaps clamp that it cannot be negative.

Image
Image

LouiseHimmelstrup:
Fixes #4971 and moves warning triangle above progress bar on the allocations page.

### DIFF
--- a/frontend-web/webclient/app/ui-components/Progress.tsx
+++ b/frontend-web/webclient/app/ui-components/Progress.tsx
@@ -8,6 +8,7 @@ import {TooltipV2} from "./Tooltip";
 import Icon from "./Icon";
 import {UNABLE_TO_USE_FULL_ALLOC_MESSAGE} from "@/Accounting";
 import {OverallocationLink} from "@/UtilityComponents";
+import Box from "@/ui-components/Box";
 
 const ProgressBaseClass = injectStyle("progress-base", k => `
     ${k} {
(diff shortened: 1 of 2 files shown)


frontend-web/webclient/app/ui-components/Progress.tsx
```diff
@@ -151,10 +150,29 @@ export function NewAndImprovedProgress({
...
+                </OverallocationLink>
+            </div>
+        </div>
+    ) : null;
+
+    return (
+        <Flex alignItems="center">
+            <div className={NewAndImprovedProgressStyle} style={style}>
+                <div className="progress-limit-overlay"/>
+                {label && <span className="progress-label">
```
Tiggles, line 169: I'd prefer if the classnames used were stored in variables (see https://github.com/SDU-eScience/UCloud/blob/master/frontend-web/webclient/app/Unstyled/index.ts#L183), but it's not important enough to block the PR.

Tiggles (approved):

DanThrane (approved):

PR #5082, opened 2025-11-21
Support cleanup

hschu12:
Changes so there should no more be any dependencies across modules


core2/pkg/orchestrator/support_assists_orc.go
```diff
@@ -23,115 +26,112 @@ func initSupportAssistsOrc() {
...
+		return orcapi.SupportAssistRetrieveProjectInfoResponse{}, err
+	}
+	jobs = foundJobs.Items
+
+	return orcapi.SupportAssistRetrieveProjectInfoResponse{
+		Project:          project,
+		ProjectWallets:   wallets,
+		ActiveGrants:     grantResults,
+		AccountingIssues: issues,
+		Jobs:             jobs,
```
DanThrane: This can go here.

DanThrane (changes requested):
I shouldn't have approved have approved the PR for this feature in the first place. That was lazy on my part. I should probably also have told you not to move on with the API design that you did send me. I was too optimistic that you would realize during development that it wouldn't work.

The API of this entire feature probably needs to be re-done. We need to change it such that the individual APIs are actually returning information from their *own* deployment and not all the others. This is a consistent error. The accounting service is returning >50% fields from the foundation. Similarly, the orchestration service is returning >50% fields from the accounting service.

PR #5155, opened 2025-12-01
Low funds scan 

### LINKED ISSUE #5030: [Core2-Accounting] Low funds check and associated email
This should possibly be done with a different algorithm for the checks.

hschu12:
Fixes #5030 
But mail does not render correctly so this is a request for flow and structure of scan algorithim.


core2/pkg/accounting/LowFundsScan.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	_, err := fndapi.MailSendToUser.Invoke(bulkRequest)
+
+	if err != nil {
+		log.Warn("Failed to LowFunds Emails", "error", err)
+		return
+	}
+
+	db.NewTx0(func(tx *db.Transaction) {
```
DanThrane: Consider skipping if there are no updates.

Issue #5162, opened 2025-12-02
[Core2-Orchestration] AppGroupID not set correct on load


hschu12:
AppGroupID is reset when core is restarted resulting in overrides of already existing app groups

DanThrane:
Duplicate of #5131

Issue #5129, opened 2025-11-27
[Core2-Orchestration] App studio browse categories order not respected


DanThrane:
Instead the categories are returned in random order, making it also impossible to re-arrange them.

Issue #5126, opened 2025-11-27
[Core2-Orchestration] Non-empty app group deletion fails


DanThrane:
Steps to reproduce:

1. Create an app group
2. Add an app to it
3. Attempt to delete the group

Issue #5125, opened 2025-11-27
[Core2-Orchestration] Crash when creating two app groups with the same name


DanThrane:
Steps to reproduce:

1. Create an application group called "mygroup"
2. Create it again

PR #5169, opened 2025-12-03
App studio issues

hschu12:
#5125 
#5126 
#5128 
#5129 
#5131 
#5162


DanThrane (changes requested):
A few things for consistency and a single bug.

DanThrane (approved):

PR #5175, opened 2025-12-08
Fixed #5092

### LINKED ISSUE #5092: [Core2-Foundation] Toggle for "allow subprojects to rename" is not functional

hschu12:


provider-integration/shared/pkg/foundation/projects.go
```diff
@@ -233,13 +247,27 @@ var ProjectToggleFavorite = rpc.Call[BulkRequest[FindByStringId], util.Empty]{
...
+
+	CustomMethod: "GET",
+	CustomPath:   "/api/" + ProjectContextV1 + "/renameable-sub",
+
+	CustomServerParser: func(w http.ResponseWriter, r *http.Request) (ProjectRetrieveSubProjectRenamingRequest, *util.HttpError) {
+		projectID := r.Header.Get("Project")
+		return ProjectRetrieveSubProjectRenamingRequest{ProjectId: projectID}, nil
+	},
+}
+
```
DanThrane: No need for a custom parser when you need to read the project from the header. The project header is automatically included in the `rpc.Actor`. As a bonus, in the new core it also correctly authenticates that the project you are sending is one that you are a member of, thus you will no longer need to do this check in the code.

The project services are a bit special in that they do not use the `Actor` (JWT) for using this authentication step, since it might be slightly out-of-date, so you will still need to keep most of your code. But you can still safely read the project ID from the actor.

```suggestion
var ProjectRetrieveSubProjectRenamingSetting = rpc.Call[util.Empty, ProjectRetrieveSubProjectRenamingResponse]{
	BaseContext: ProjectContextV1,
	Convention:  rpc.ConventionQueryParameters,
	Roles:       rpc.RolesEndUser,
	Operation:   "renameable-sub",
}
```

DanThrane (changes requested):
Added some tips for using all the features of the new Core. The reason that the toggle keeps the project ID from the request is because it comes from the request body rather than the header.

PR #5176, opened 2025-12-09
Fixes #5132

### LINKED ISSUE #5132: [UserInfo] Extend visibility of user-info screen
## About the feature

Currently, this feature is shown only on the user settings page and is hidden behind the `Feature.ADDITIONAL_USER_INFO`. I have been told that we can go ahead and remove the feature-flag and extend the functionality. The current widget allows users to enter in additional information about themselves.

Image

## When should the widget be shown?

The component should not be more visible and be shown to users when certain actions _if it hasn't been filled out already_. This includes in the following cases:

1. **After successfully logging in.** This screen should be optional. Submitting the form should take you to the dashboard.
2. **After joining a project via a link.** This screen should be optional. Submitting the form should take you to the dashboard.
3. **When applying for new resources.** Should be mandatory and be embedded in the application form. The green save button should _not_ be visible and instead it should automatically be submitted along with the application. This part of the form should then automatically disappear when viewing the submitted application.

## Extension of UI

In addition to displaying the UI in more places, it should be extended to give concrete options instead of relying purely on free-text fields. This will be a frontend only feature, since free-text fields will still be used as a fallback in cases where we are unable to give concrete options.

**Full name of organization:** This field should be mandatory. Fill this based on the orgId present in the JWT. This should do a mapping based on the organization that we know. For example, this might map "sdu.dk" to "University of Southern Denmark". If an organization is known, then this field should be disabled. If it is not known, then it should be displayed as a free-text field like it is today.

**Department:** This field should be mandatory. If the organization is not known, then this should show a free-text field for the department. Assuming that the organization is known, this should show a hierarchical dropdown of options that we know. For example, for SDU it might show:

- Faculty of Science:
  - Department of Biology
  - Department of Biochemistry and Molecular Biology
  - Department of Physics, Chemistry and Pharmacy
  - Department of Mathematics and Computer Science
- Faculty of Humanities
  - List of departments...
- Faculty of Business and Social Sciences
   - List of departments...
- Faculty of Health Sciences
   - List of departments...
- Faculty of Engineering
   - List of departments...

Selecting "Department of Mathematics and Computer Science" in this case should result in a value of `"Faculty of Science/Department of Mathematics and Computer Science"` for the backend.

For known organizations that are not universities this can be treated as free-text field.

**Position:** This field should be mandatory. Dropdown of common university positions. Should make sure to include at least positions from:

Simple version:

- VIP
- TAP
- Student

~Old version:~

- ~Administration (could be kept generic)~
- ~Professors (of various levels)~
- ~PhD student (probably fine as it is)~
- ~Research assistants and similar roles~
- ~Technical staff (maybe fine)~
- ~Various IT roles (a few different roles?)~

**Research field:** This field should be mandatory. It should contain items from [this list](https://github.com/user-attachments/files/23815922/oecd-research-areas.txt).

## Dashboard

In addition to the changes mentioned above, a small Grafana dashboard should be developed for showing various statistics. This should be based on SQL queries made against the database. The following tables are relevant for gathering information:

- `auth.principals`: Generic user-information. Most importantly this also contains information about their organization (`org_id`).
- `auth.additional_user_info`: Table containing information from the above mentioned UI widget.

The dashboard should aim to give a breakdown of how the platform is used by different organizations. This part is up to interpretation, but should probably include at least:

1. Global breakdown of position
2. Global breakdown of research field
3. By-organization breakdown of position/research field
4. Breakdown between departments at an organization

## Known organizations

The widget should be able to handle all danish universities along with other organizations which we have partnerships with.

Tiggles:
Basic general example of usage
https://github.com/user-attachments/assets/750398f0-e72b-41ad-8c9d-f185f3983968

Basic general example of usage with provided known org id
https://github.com/user-attachments/assets/c6902ab9-23f3-44f2-b1ef-ecb15775ec11

Basic general example of usage with provided unknown org id
https://github.com/user-attachments/assets/65ec2025-8f9f-49b8-8f99-020c83eeabc0

Modal shown on login action
https://github.com/user-attachments/assets/da849846-fd75-4ebb-a82c-50037101fc1e

Modal shown on project invite accept
https://github.com/user-attachments/assets/3a16b660-426f-4b96-9a82-157863090156

Applying for resources.
Note: This will always be shown as gender field is always empty when fetching info from backend. Should be resolved when this is added to the optionalUserInfo backend.
Organization info isn't shown when viewing existing grants.
https://github.com/user-attachments/assets/0eea0f80-edba-4aa9-a03d-8eb2fbe7b74b

### DIFF
--- a/backend/auth-service/api/src/main/kotlin/UserDescriptions.kt
+++ b/backend/auth-service/api/src/main/kotlin/UserDescriptions.kt
@@ -87,6 +87,7 @@ data class OptionalUserInformation(
     val department: String? = null,
     val researchField: String? = null,
     val position: String? = null,
+    val gender: String? = null,
 )
 
 object UserDescriptions : CallDescriptionContainer("auth.users") {
--- a/backend/auth-service/src/main/kotlin/services/PrincipalService.kt
+++ b/backend/auth-service/src/main/kotlin/services/PrincipalService.kt
@@ -659,7 +659,8 @@ class PrincipalService(
                         info.organization_full_name,
                         info.department,
                         info.research_field,
-                        info.position
+                        info.position,
+                        info.gender
                     from
                         auth.principals p join
                         auth.additional_user_info info on p.uid = info.associated_user
@@ -677,6 +678,7 @@ class PrincipalService(
                     row.getString(1),
                     row.getString(2),
                     row.getString(3),
+                    row.getString(4)
                 )
             }
         }
@@ -695,18 +697,20 @@ class PrincipalService(
                     setParameter("department", info.department)
                     setParameter("research_field", info.researchField)
                     setParameter("position", info.position)
+                    setParameter("gender", info.gender)
                 },
                 """
                     insert into auth.additional_user_info (associated_user, organization_full_name, department,
-                        research_field, position) 
-                    select p.uid, :organization_full_name, :department, :research_field, :position
+                        research_field, position, gender) 
+                    select p.uid, :organization_full_name, :department, :research_field, :position, :gender
                     from auth.principals p
                     where p.id = :username
                     on conflict (associated_user) do update set
                         organization_full_name = excluded.organization_full_name,
                         department = excluded.department,
                         research_field = excluded.research_field,
-                        position = excluded.position
+                        position = excluded.position,
+                        gender = excluded.gender
                 """
             )
         }
--- a/backend/auth-service/src/main/resources/db/migration/V29__Gender.sql
+++ b/backend/auth-service/src/main/resources/db/migration/V29__Gender.sql
@@ -1,0 +1,1 @@
+alter table auth.additional_user_info add column if not exists gender text;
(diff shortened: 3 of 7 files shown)


Tiggles:
I'm not sure I understand this one, or I can't reproduce it.
> Pressing "enter" in the data list without the modal shown will still trigger the suggestion acceptance.

This is current behavior, isn't it?
> When free-text mode is allowed, then I think we should use the text cursor when hovering the field. In other words, it should feel like a completely normal <input>

Is this just adding a chevron or swap the component (<Input />) with a `<Select/>`/`<RichSelect/>` when options are fixed?
> When free-text is not allowed, the data list should probably show an icon indicating a dropdown menu. I think it would make sense to style it similar to a <select>. We probably want to use the "pointer" cursor here since we use that it most other places.

Most bugs and consistency are fixed, still working on some.

DanThrane:
> I'm not sure I understand this one, or I can't reproduce it.

I think you have fixed this by fixing the other tab button. The modal I was referring to was the select options. The bug was that you could trigger one of the suggestions even though the modal wasn't open. So my guess is that this is has gone away now that the modal correctly opens when pressing on focus.

> This is current behavior, isn't it?

Looks like it. I might have been confused about the non free-text mode actually. I just checked your branch and it looks correct.

> Is this just adding a chevron or swap the component () with a <Select/>/<RichSelect/> when options are fixed?

Whichever is easier. I would imagine that adding the chevron is easier.

DanThrane (approved):
Looks good to me. I will merge this pending #5197.

PR #5198, opened 2025-12-10
fixes #5150 and fixes #5130

### LINKED ISSUE #5130: [Core2-Orchestration] Impossible to activate a new spotlight
Steps to reproduce:

1. Create a new spotlight
2. Attempt to activate it from the UI

### LINKED ISSUE #5150: [Core2-Orchestration] App store content disappears when deleting active spotlight
When deleting the active spotlight and reloading the page, everything except the starred applications disappear from the applications page.
The content returns upon restarting the core, with the spotlight being deleted correctly.  Cache and the database may be out of sync

hschu12:
#5150 did not happen after changes that fixed #5130

Have attempted different things to make the frontend reload the landing page on spotlight update but without succes. The page needs to be refreshed after new spotlight is applied for it to show/change.

### DIFF
--- a/core2/pkg/orchestrator/app_catalog.go
+++ b/core2/pkg/orchestrator/app_catalog.go
@@ -1254,6 +1254,11 @@ func AppStudioListSpotlights() []orcapi.Spotlight {
 		}
 	}
 
+	// Henrik: Sort by titles
+	sort.Slice(result, func(i, j int) bool {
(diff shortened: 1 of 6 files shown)


DanThrane (approved):
I would have preferred not to call it `Toggle` given that it is not doing a toggle but rather a store/set. But this is small enough that I can just change it.

Issue #5214, opened 2025-12-16
[Core2-Foundation] News - listing same date news are ordered wierdly when hidden is selected and unselected


hschu12:
Create multiple news starting and ending at the same time (perhaps same category)
Click the hidden button and see it move, when unselected it does not move back og if multiple are hidden they usally are moved

Issue #5213, opened 2025-12-16
[Core2-Foundation] News can only show one news with same start time


hschu12:
Create two news with same start and end time and  reload dashboard
The news section just shows a random instaed of both.

DanThrane:
I would argue that it shouldn't attempt to show both but rather show the last one that was created (highest id). I agree that it definitely shouldn't show a random one.

Issue #5212, opened 2025-12-16
[Core2-Foundation] News - Categories are case sensitive but not shown as it


hschu12:
Most likely fixed easiest in the backend just make it case sensitive

Created multiple posts with category/tag "FLOW" and "flow"

- When shown as the tag it is all caps
- In URL it is case sensitive (so correct)
- When list is filtered by tag it is always with capital start letter followed by lowercases
- When creating new news it lists the existing categories both as Flow when in fact it is FLOW and flow

Issue #5211, opened 2025-12-16
[Core2-Foundation] News - Cannot select category on news/list


hschu12:
From dashboard select a news story
Click the tag
unselect the tag
You cannot select a tag from the list of news

PR #5227, opened 2025-12-17
News issues

hschu12:
#5187 
#5211 
#5212 
#5213 
#5214


core2/pkg/foundation/news.go
```diff
@@ -106,9 +107,13 @@ func (row *newsRow) ToApi() fndapi.NewsPost {
...
 
 func NewsBrowsePosts(request fndapi.ListPostsRequest) (fndapi.Page[fndapi.NewsPost], *util.HttpError) {
-	request.ItemsPerPage = 50
 
 	return db.NewTx(func(tx *db.Transaction) fndapi.Page[fndapi.NewsPost] {
+		categoryFilter := request.Filter.GetPtrOrNil()
+		if categoryFilter != nil {
+			filterString := strings.ToUpper(*categoryFilter)
+			categoryFilter = &filterString
+		}
```
DanThrane: No need to use `GetPrOrNil()` on an option in this case. Just use the `Present` and `Value` fields. This method is mainly there for 3rd party libraries which expect a pointer when passing things that might not be present. We rarely, if ever, needed it within our own code.

PR #5234, opened 2025-12-19
Fixes #5147

### LINKED ISSUE #5147: [Core2-Orchestration] Permissions modal and permissions page for public IPs should not display read/write/none permissions
The page should display the same options as when an IP is created

Tiggles:


DanThrane (changes requested):
This just turns off the permissions, it is supposed to show the permission page which is present when you create the IP.

PR #5295, opened 2026-01-28
Added migration for visits and logic for handling unread comments in …

vushu:
What has been made:

- Added migration for tracking user visits on grant applications
- Added logic for handling the unread comments
- Tested using different users.

This is only the backend, no frontend has been implemented yet.

The GrantStatus struct now has the hasUnreadComments, 
which the frontend can use for indicating unread comments.

```
type GrantStatus struct {
	OverallState      GrantApplicationState     `json:"overallState"`
	StateBreakdown    []GrantGiverApprovalState `json:"stateBreakdown"`
	Comments          []GrantComment            `json:"comments"`
	Revisions         []GrantRevision           `json:"revisions"`
	ProjectTitle      util.Option[string]       `json:"projectTitle"`
	ProjectPI         string                    `json:"projectPI"`
	HasUnreadComments bool                      `json:"hasUnreadComments"`
}
```


core2/pkg/accounting/grants.go
```diff
@@ -1064,6 +1067,81 @@ func GrantsUpdateState(actor rpc.Actor, req accapi.GrantsUpdateStateRequest) *ut
...
+			do update set last_visited_at = excluded.last_visited_at;`,
+			db.Params{
+				"application": applicationId,
+				"user":        username,
+			},
+		)
+	})
+}
+
+func grantRetrieveUserApplicationVisits(applicationId string, username string) time.Time {
```
DanThrane: Any chance that we can change this function such that it primarily reads from an in-memory cached version? I am a bit worried that this will end up doing N database queries per application returned. The code itself looks fine, but I would probably just change it a bit:

- Add a new type of bucket for data associated with a specific user (in `grantGlobals`)
- Find the appropriate bucket for the user, using the username as the key.
- The bucket itself should be protected by an RW mutex. Similar to the others. Attempt to read the data from the bucket, if not present, switch to a write lock and fetch the data as you already do, but drop the `application_id` filter.

I can help you out with modifying the code.

core2/pkg/accounting/grants.go
```diff
@@ -1064,6 +1111,82 @@ func GrantsUpdateState(actor rpc.Actor, req accapi.GrantsUpdateStateRequest) *ut
...
 }
 
+// Grant Application has unread comments
+// =====================================================================================================================
+// This section handles unread comments inside applications.
+// Firstly, we record user visits on GrantRetrieve function
+// When GrantApplicationProcess is triggered, we will then compare the last time the user
+// has visited the application with the latest comment of the given application, if the latest comment is greater,
+// then we will set app.Status.HasUnreadComments = true
+func grantUserHasUnreadComments(app *accapi.GrantApplication, username string) {
```
DanThrane: I would change the name of this function to make it clear that this is not answering a query, but rather attaching the metadata.

Probably something like `grantAttachUnreadCommentStatus` would do the trick.

DanThrane: I would an empty line before this. Go will assume that the comment belongs to the function otherwise. In this case it more belongs to the section.

DanThrane (dismissed):

PR #5297, opened 2026-01-30
Changed error handling for read-only drives to unauthorized instead o…

vushu:
I changed it for unauthorized instead of not found.


vushu:
I need to further investigate it since after testing it wasn't resolved.

PR #5298, opened 2026-01-30
Changed limit to percentage and also fixed errors in flow

hschu12:


DanThrane (commented):
I am not a 100% certain that this code correctly works with retiring allocations. Can you check that these values are actually correct? Can you also remind me if there is any specific reason we are choosing to use read from the database instead of from the in-memory version? We might have discussed this once before, but in that case I forgot.

DanThrane (approved):

PR #5345, opened 2026-02-09
Issue 5001

vushu:
Added bleve search for the grants page

Currently search indexes
```
CreatedBy      []string
RecipientTitle []string
RecipientId    []string
ReferenceIds   []string
Comments       []string
```

### DIFF
--- a/core2/pkg/accounting/grants.go
+++ b/core2/pkg/accounting/grants.go
@@ -1248,7 +1248,8 @@ func grantSearchFuzzily(searchTerm []string) []string {
 
 	res, err := grantIndex.Search(req)
 	if err != nil {
-		panic(err)
+		log.Warn("Failed at searching for grants: %s", err)
+		return []string{}
 	}
 
 	var results []string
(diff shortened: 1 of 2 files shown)


frontend-web/webclient/app/Grants/GrantApplicationBrowse.tsx
```diff
@@ -82,25 +83,41 @@ export function GrantApplicationBrowse({opts}: {opts?: ResourceBrowserOpts<Grant
...
                     columns[3].columnWidth = 0;
                 }
-                browser.setColumns(columns);
-                browser.on("open", (oldPath, newPath, resource) => {
-                    if (resource) {
-                        navigate(AppRoutes.grants.editor(resource.id));
-                        return;
-                    }
 
+                function getGrants(newPath?: string) {
```
Tiggles: Not something we have discussed at any point, but I prefer `fetchGrants`, as we usually (but probably not always) use `fetch` when doing a network call and `get` when is done locally from a cache.

PR #5376, opened 2026-02-19
Issue 5340

vushu:
Added optional user info to grant status and grant export to csv

### DIFF
--- a/core2/pkg/accounting/grants_export.go
+++ b/core2/pkg/accounting/grants_export.go
@@ -103,9 +103,9 @@ func GrantsExportBrowseToCsv(lines []accapi.GrantsExportResponse) string {
 		}
 	}
 
-	categoriesArray := []string{}
+	var categoriesArray []string
 	for item := range categories {
-		categoriesArray = append(categoriesArray, strings.ReplaceAll(item, "-", "_"))
+		categoriesArray = append(categoriesArray, item)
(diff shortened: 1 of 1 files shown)


core2/pkg/foundation/auth_user_info.go
```diff
@@ -7,13 +7,14 @@ import (
...
 	fndapi "ucloud.dk/shared/pkg/foundation"
+	"ucloud.dk/shared/pkg/log"
 	"ucloud.dk/shared/pkg/rpc"
 	"ucloud.dk/shared/pkg/util"
 )
 
-func UserOptInfoRetrieve(actor rpc.Actor) fndapi.OptionalUserInfo {
-	result := db.NewTx(func(tx *db.Transaction) fndapi.OptionalUserInfo {
-		row, _ := db.Get[struct {
+func UserOptInfoRetrieve(username string) (fndapi.OptionalUserInfo, bool) {
```
DanThrane: This was intentionally returning a default object in case nothing was ever put in the database. There should probably have been a comment, but I think it makes sense to just return the default object (which has all properties set to null).

DanThrane (changes requested):
Mostly looks fine, but I would avoid changing the semantics of the `UserOptInfoRetrieve` rpc.

Issue #5350, opened 2026-02-09
[Notifications] Opt-in sounds for notifications


Tiggles:
Not sure how popular this feature would be or if people are actively against it, but we could at notification sounds that play if, and only if, the user has opted in to them.

This could happen when the user is waiting for a job to start and the tab plays a sound when the job transitions to running, instead of having to repeatedly check the tab.

Other notifications could also have sounds, but I'm not sure which actually make sense to have it for.

DanThrane:
This project is slightly broader than what is mentioned here, but I still think this project looks interesting for something like this: https://github.com/Danilaa1/cuelume

Issue #5346, opened 2026-02-09
Notification changes


Tiggles:
General changes that should be made for notifications and something to discuss

PR #5441, opened 2026-03-12
Issue #5346

Tiggles:
Fixes sub-issues of #5346, with the exception of #5350


DanThrane (changes requested):
Looks pretty good to me, I found a few issues worth mentioning.

PR #5474, opened 2026-03-20
Issue 5413

vushu:
Audit logging feature implementation.


DanThrane (changes requested):
These two are admittedly not a big issue in practice.

DanThrane (approved):

PR #5546, opened 2026-04-27
Grant changes

hschu12:
Makes it possible to 
- See the grant related to personal workspace gifts. Before it would not show since the system creates the gift and we looked at createdBy. Now we check if it is to er personal workspace and then checks if the recipient is the user.
- Gift display is now shown to providers as before, but also to any project that is a sub-allocator (cannot consume resources)

Also changed the look from `Accordion` to   use `Tree` and `TreeNode` so that it indents when expanded. Since this also resulted in `Accordion` not being used anywhere anymore it has now been deleted from the code base.


frontend-web/webclient/app/UCX/UcxView.tsx
```diff
@@ -120,25 +120,25 @@ const UcxAccordion: React.FunctionComponent<React.PropsWithChildren<{
...
+                                                            sysHello,
+                                                            maxReconnectAttempts = Number.POSITIVE_INFINITY,
+                                                            renderFrame,
+                                                            components,
+                                                            functions,
+                                                            rpcHandlers,
+                                                            onConnected,
+                                                            onDisconnected,
+                                                            onTransportError,
+                                                        }) => {
```
Tiggles: I think this can be less wide

frontend-web/webclient/app/Accounting/Allocations/index.tsx
```diff
@@ -305,13 +306,21 @@ const Allocations: React.FunctionComponent = () => {
...
+                <Box flexGrow={1}/>
+                <ProjectSwitcher/>
             </header>
 
-            <ProviderOnlySections state={state} dispatchEvent={dispatchEvent} />
+            {(state.remoteData.managedProviders ?? []).length > 0 ? <>
+                <RootAllocationSections state={state} dispatchEvent={dispatchEvent}/>
+                <GiftSection state={state} dispatchEvent={dispatchEvent}/>
+            </> : <>
+                {checkCanConsumeResources(Client.projectId ?? null, {api: FilesApi}) ? null :
```
DanThrane: That is not the criteria the backend uses. The backend explicitly checks if you are the provider project.

I agree that we need to make it slightly more broad, but it cannot be so broad that everybody which has `!canConsumeResources` is allowed to do it. All users are capable of creating such a project. We need to put strict controls on who can create these gifts to avoid end-users "maliciously" spamming people with allocations.

DanThrane (changes requested):
Whitespace changes in too many files that shouldn't be touched at all. I disagree with the criteria for showing the gift section because it is too broad. It does make it appear when it should, but it makes it appear in too many cases.

I am okay with any solution which has gifts on an UCloud admin controlled allow-list.

Tiggles:
If this criteria is now considered correct, the rest of the front-end seems fine.

> I disagree with the criteria for showing the gift section because it is too broad. It does make it appear when it should, but it makes it appear in too many cases.

vushu (approved):

PR #5590, opened 2026-05-26
Fix race condition

### LINKED ISSUE #5527: [UI] Hide resource browser results for allocator projects
When a project cannot consume resources, reloading the page can still show results that should not be available.

**Relevant source locations:**

- `frontend-web/webclient/app/ui-components/ResourceBrowser.tsx`

**Success criteria:**

- The browser respects `canConsumeResources` after reload.
- Unavailable projects do not show consumable results.
- Cached project data does not bring the results back.

Tiggles:
Handle session filters still being shown when canConsumeResources === false

Fixes #5527


hschu12 (approved):
If it is tested then I do not see any problems

PR #5594, opened 2026-05-28
Fix job/view file redirection

Tiggles:


hschu12 (approved):
If tested and works, then I see no code releveant issueus

PR #5639, opened 2026-07-01
Issue 5526 

Tiggles:
Changes to the product selector. For non-compute, the provider has been added. As the width of the selector has expanded, this is centered for this case:

Screenshot 2026-07-01 at 14 10 29

For compute:

The selection of a compute product has been divided into three parts:

- Service provider
- Compute category (grouping of product types)
- A slider to select specific product category

Screenshot 2026-07-01 at 14 27 31

Screenshot 2026-07-01 at 14 13 50

Note: The above product type has no GPUs, so the product stat does not show anything regarding GPUs.

The service provider list is populated from the available products, and the first available is selected, if any. On workspace change, the selected service provider persists iff the user has at least one product available at this provider, otherwise, the first available is selected.

If no service provider is available for the workspace, the service provider box is greyed out, but the machine selection is not. The "Apply for resources" button is available here.
Screenshot 2026-07-01 at 14 18 31

If the compute category has both fractional and non-fractional resources (e.g. MIGs/Full GPUs), a divider is shown:

Screenshot 2026-07-01 at 14 20 14

Here, the product stats shows '1 MIG(s)', and would show '`n` GPU(s)' if selecting a product type to the right of the divider.

The short description for the product category is shown in both the dropdown and above the product stats. The machine status is also written out, with a background corresponding to it's queue status, i.e. green, yellow and red. Finally (?), a GPU-icon has been added.

I have spoken with Claudio, Federica, and Jonas wrt. feedback.


DanThrane (commented):
I have no problems with the code, but I do have a few stylistic comments. Should all be fairly simple.

Image

Image

Tiggles:
Screenshot 2026-07-02 at 12 12 53
Screenshot 2026-07-02 at 12 12 21

DanThrane (approved):

PR #5653, opened 2026-07-06
A bunch of selector fixes

Tiggles:
Description should textSecondary
Product name margin shouldn't be there
Clicking the table wrongly opens the machine type selector
Clicking the numbers below the input range should jump to that option
Track color should be gray-ish (also divider)

Screenshot 2026-07-06 at 14 12 51
Screenshot 2026-07-06 at 14 12 43

Not sure about the gray color choice.


Tiggles:
I didn't move the product description, as you can see in the images.

DanThrane (approved):

PR #5741, opened 2026-08-19
Issue 5691/connect using ucloud cli

vushu:
Initial implementation of  connect command for ucloud-cli


frontend-web/webclient/app/Login/ExternalLogin.tsx
```diff
@@ -25,13 +25,17 @@ interface ExternalLoginService {
 
 const externalLoginServices: ExternalLoginService[] = [
     {
-        id: "test",
-        title: "Test service",
+        id: "ucloud-cli",
+        title: "UCloud-cli service",
         description: "Send a new UCloud session to the development console.",
```
DanThrane: Should probably tweak this at some point.

frontend-web/webclient/app/Login/ExternalLogin.tsx
```diff
@@ -25,13 +25,17 @@ interface ExternalLoginService {
...
-        feature: Feature.EXTERNAL_LOGIN_TEST,
+        feature: Feature.EXTERNAL_LOGIN_UCLOUD_CLI,
         handle(response) {
-            // eslint-disable-next-line no-console
+            const redirectUrl = new URL("http://localhost:59421/auth");
+            redirectUrl.searchParams.set("username", response.username);
+            redirectUrl.searchParams.set("token", response.refreshToken);
             console.log("External login response", response);
```
DanThrane: Remove

ucloud-cli/pkg/command/connect.go
```diff
@@ -1,16 +1,191 @@
...
 func (c *ConnectCommand) Execute() error {
-	return fmt.Errorf("connect not implemented")
+	return performConnection(c.Dev)
+}
+
+func getRoot(w http.ResponseWriter, r *http.Request) {
+	io.WriteString(w, "UCloud-CLI server is running!")
+}
+
+func successMessage() string {
```
DanThrane: We may want to style this further at some point.

ucloud-cli/pkg/command/connect.go
```diff
@@ -1,16 +1,191 @@
...
+	"net/http"
+	"os/exec"
+	"runtime"
+	"time"
+
+	"ucloud.dk/ucloud_cli/pkg/shared"
+)
+
+const PORT = ":59421"
+const UCLOUD_CLI_PATH = "/app/login/external?service=ucloud-cli"
```
DanThrane: Go convention is to just apply normal naming even to constants. You can also do multiple in one statement. Formatting is a bit weird in GitHub's code review, so I can't quite align them.

```suggestion
const (
	Port = ":59421"
	UCloudCliPath  = "/app/login/external?service=ucloud-cli"
)
```

ucloud-cli/pkg/command/connect.go
```diff
@@ -1,16 +1,191 @@
...
 
 func (c *ConnectCommand) Execute() error {
-	return fmt.Errorf("connect not implemented")
+	return performConnection(c.Dev)
+}
+
+func getRoot(w http.ResponseWriter, r *http.Request) {
+	io.WriteString(w, "UCloud-CLI server is running!")
+}
+
```
DanThrane: Could probably be removed.

ucloud-cli/pkg/command/connect.go
```diff
@@ -1,16 +1,191 @@
...
+		return fmt.Errorf("start auth server: %w", err)
+	}
+	defer ln.Close()
+
+	server = &http.Server{
+		Handler:           mux,
+		ReadHeaderTimeout: 5 * time.Second,
+	}
+
+	ready <- struct{}{}
```
DanThrane: We sometimes use `util.Empty` here, but honestly this is also fine.

ucloud-cli/Makefile
```diff
@@ -0,0 +1,18 @@
+.DEFAULT_GOAL := build
```
DanThrane: Just to keep it consistent with the rest of the code base, I think we should use just a simple shell script (see other build.sh and other service.sh scripts). The service scripts are usually the ones that are used for development (the launcher invokes them, but we used to do it by hand).

ucloud-cli/pkg/shared/config.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func SaveConfig(cfg *Config) error {
+	data, err := yaml.Marshal(&cfg)
+	if err != nil {
+		return err
+	}
+	fmt.Println("Saving config to:", GetConfigPath())
+
+	return os.WriteFile(GetConfigPath(), data, 0600)
```
DanThrane: I would change this to write to a temporary file first and only after it succeeds replace the actual config file via a move. This should make sure we never end up in a state of having a partially written file.

ucloud-cli/pkg/shared/config.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	return &cfg, nil
+}
+
+func SaveConfig(cfg *Config) error {
+	data, err := yaml.Marshal(&cfg)
+	if err != nil {
+		return err
+	}
+	fmt.Println("Saving config to:", GetConfigPath())
```
DanThrane: These should go away at some point.

ucloud-cli/pkg/command/connect.go
```diff
@@ -1,16 +1,191 @@
...
+
+	}
+	currentEnv := cfg.Environments[cfg.DefaultEnvironment]
+	baseURL := currentEnv.URL
+	if dev {
+		baseURL = shared.DEV_SERVER
+	}
+
+	go func() {
+		if err := startAuthServer(ready, authDone); err != nil {
```
DanThrane: This server is never cancelled in case any non server error is hit. Could probably cause some weird cases where there application hangs.

DanThrane (dismissed):
A few minor comments, nothing too serious. I don't mind letting it be merged now and let you address the comments you find helpful at some point later when it makes sense.

ucloud-cli/pkg/shared/config.go
vushu: Good point.

DanThrane (approved):

PR #5745, opened 2026-08-24
Some UI cleanup, regarding #5712

### LINKED ISSUE #5712: [UI] UI cleanup
Epic to track various responsiveness, consistency and correctness bugs in the UI.

Tiggles:


frontend-web/webclient/app/Files/SharesOutgoing.tsx
```diff
@@ -58,9 +58,11 @@ let RIGHTS_TOGGLE_ICON_CACHE: {
     REJECTED: null,
 }
 
-new ReactStaticRenderer(() =>
-    <RadioTilesContainer height={48} onClick={stopPropagation}>
+new ReactStaticRenderer(() => {
+    const foo = Math.random().toString();
```
DanThrane: This doesn't solve the issue. The result of `ReactStaticRenderer` is being cloned later in this file so you are just copying the same randomly generated ID.

frontend-web/webclient/app/Applications/Category.tsx
```diff
@@ -8,35 +8,22 @@ import {useAppSearch} from "./Search";
...
+import {NoResultsBody} from "@/UtilityComponents";
 
 const ApplicationsCategory: React.FunctionComponent = () => {
     const location = useLocation();
     const idParam = getQueryParam(location.search, "categoryId");
     const id = parseInt(idParam ?? "-1");
 
     const [discovery] = useDiscovery();
-    const [categoryState, fetchCategory] = useCloudAPI(AppStore.retrieveCategory({id, ...discovery}), null);
+    const [categoryState, fetchCategory] = useCloudAPI<AppStore.ApplicationCategory | null>({noop: true}, null);
```
DanThrane, line 26: This is more a note of something we might want to consider changing more globally. In this case the spinner will not show on the initial render and the "No applications found" message will show since on the initial render `loading = false` and `groups.length === 0`. Only on the second render when `refresh()` has fired does `loading` actually become `true`. I am sure we probably have loads of these scattered around the codebase. Really just a foot-gun from `useCloudAPI()`. I think we can probably fix this by just setting it to initially have `loading: true`?  Let me know what you think.

frontend-web/webclient/app/Applications/Category.tsx
```diff
@@ -62,7 +55,9 @@ const ApplicationsCategory: React.FunctionComponent = () => {
                         <Box ml="auto" />
                         <UtilityBar onSearch={appSearch} />
                     </Flex>
-
+                    {categoryState.loading ? <HexSpin size={128} />:
+                            groups.length !== 0 ? null :
+                                <NoResultsBody title={`No applications found category`} children={undefined} />}
```
DanThrane:
```suggestion
                                <NoResultsBody title={"No applications found in category"} />}
```

I think you forgot a word.

frontend-web/webclient/app/Files/SharesOutgoing.tsx
Tiggles: It was a mistake to change here, as it is supposed to be overridden elsewhere. There was a different bug with that approach, but should be fixed in 9091828caec398ee932c1141f1d57c1b26529799, line 404 to 406.

frontend-web/webclient/app/Applications/Category.tsx
Tiggles: Seems fine. My thoughts are that `useCloudAPI` is usually to fetch resources on initial load, so defaulting to `true` makes sense, while `callAPI` handles delayed effects/calls, that shouldn't be affected by this change.

DanThrane (approved):

PR #5748, opened 2026-08-26
Usage issues

hschu12:
Created code for usage generation of Capacity based products that can increase and decrease in a normal usage pattern.

Changed code to collapse reports to be more correct and fill in empty timeslots.

Changes to Usage Page

General:
- All charts show same top 10 projects based on highest absolute usage at end of time range
- Uses same color coding across charts

Pie chart:
- Reintroduced Pie chart of sub project usage. 
- Shows absolute usage at latest point in select time range both for GB and for compute.
- Changed labeling to only show if can be contained within the slice

Screenshot 2026-08-26 at 12 17 44

Usage over time:
- Stacked.  Shows the usage of each top 10 project by stacking usage.
- Seperated. shows the usage of each top 10 project by individual lines.
Screenshot 2026-08-26 at 12 19 50
Screenshot 2026-08-26 at 12 19 59

Change in usage over time:
- Does not show on storage since this does not make sense for negative numbers
- Shows how much each of the top 10 projects has used in each bucket
Screenshot 2026-08-26 at 12 20 15

Utilization:
- Minor bug fix resulting in removal of double spikes in utilization or usage due to overlapping reports.
Screenshot 2026-08-26 at 12 20 09


frontend-web/webclient/app/Accounting/Diagrams/DeltaOverTime.tsx
```diff
@@ -71,54 +71,67 @@ export function useDeltaOverTimeChart(
...
-            for (const [childKey, childEntries] of Object.entries(byChild)) {
-                const change = childEntries.reduce((acc, e) => acc + e.change, 0);
-                summed.push({ timestamp: ts, child: childKey === "" ? null : childKey, change });
-            }
-        }
+        // Split stack into positive and negative
+        const keys = (union(data.map(it => it.child ?? "")))
 
-        data = summed;
+        const positiveStack = stack<number>()
```
DanThrane, line 93: `positiveStack` and `negativeStack` are identical?

core2/pkg/accounting/accounting_usage_generator.go
```diff
@@ -3,25 +3,51 @@ package accounting
...
+
+const (
+	UsageGenProductCPU UsageGenProduct = iota
+	UsageGenProductStorage
+)
+
 type UsageGenApi struct {
-	AllocateEx  func(now, start, end int, quota int64, recipientRef, parentRef string)
-	ReportDelta func(now int, ownerRef string, usage int64)
+	AllocateEx  func(product UsageGenProduct, now, start, end int, quota int64, recipientRef, parentRef string)
```
DanThrane, line 22: Test is broken due to API changes.

core2/pkg/accounting/usage_report.go
```diff
@@ -485,6 +521,7 @@ func usageRetrieveHistoricReports(from time.Time, until time.Time, wallet AccWal
 }
 
 func usageCollapseReports(reports []internalUsageReport) internalUsageReport {
+	fmt.Printf("Collapsing reports: \n")
```
DanThrane: Log statement.

core2/pkg/accounting/usage_report.go
```diff
@@ -518,27 +555,212 @@ func usageCollapseReports(reports []internalUsageReport) internalUsageReport {
...
+	// Delta data grouped by child (for later)
+	deltaByChild := map[AccWalletId]map[time.Time]int64{}
+	allDeltaTimestamps := map[time.Time]util.Empty{}
+
+	// Absolute usage for local
 	absoluteUseByTimestamp := map[time.Time]int64{}
 	absoluteQuotaByTimestamp := map[time.Time]int64{}
 
 	for _, report := range reports {
+		fmt.Printf("report being processed: %v\n", report)
```
DanThrane: Log statement.

core2/pkg/accounting/usage_report.go
```diff
@@ -1149,41 +1359,56 @@ func lUsageSampleWallet(now time.Time, cmp internalSnapshotComparison, b *db.Bat
...
 
-				parentReport.Dirty = true
-			} else {
-				parentReport.SubProjectHealth.Idle++
-			}
+		parentReport.UsageOverTime.ChildrenAbsolute = append(
+			parentReport.UsageOverTime.ChildrenAbsolute,
+			internalUsageOverTimeAbsoluteChildrenDataPoint{
+				Timestamp: now,
+				Usage:     currWallet.TotalUsage,
```
DanThrane: Is this the correct value to use? Won't this tell all parents about the wallet's combined usage?

core2/pkg/accounting/usage_report.go
```diff
@@ -295,6 +323,7 @@ func initUsageReports() {
 
 			var reports []accapi.UsageReport
 			for _, w := range wallets {
+				fmt.Printf("wallet: %v\n", w)
```
DanThrane: Log statement.

core2/pkg/accounting/usage_report.go
```diff
@@ -307,11 +336,14 @@ func initUsageReports() {
 				endTime := fndapi.TimeFromUnixMilli(request.End).Time()
 
 				historicReports := usageRetrieveHistoricReports(startTime, endTime, walletId)
-
+				for _, report := range historicReports {
+					fmt.Printf("report: %v \n", report)
+				}
```
DanThrane, line 392: Looped log statement.

core2/pkg/accounting/usage_report.go
```diff
@@ -112,10 +123,20 @@ func (r *internalUsageOverTimeDeltaDataPoint) ToApi() accapi.UsageReportDeltaDat
...
-		Timestamp: fndapi.Timestamp(r.Timestamp),
-		Change:    r.Change,
+type internalUsageOverTimeAbsoluteChildrenDataPoint struct {
+	Timestamp time.Time
+	Usage     int64
+	Child     util.Option[AccWalletId]
+}
+
+func (d *internalUsageOverTimeAbsoluteChildrenDataPoint) ToApi() accapi.UsageReportAbsoluteChildrenDataPoint {
+	child := setChild(d.Child)
```
DanThrane, line 171: This happens late enough in the process that if we are collapsing multi product category reports (that share the same unit), then we can end up with duplicate rows such as:

```json
[
  {"timestamp": 1, "child": "project-a", "usage": 100},
  {"timestamp": 1, "child": "project-a", "usage": 100}
]
```

In the collapse report function it is done by wallet ID. But the same product can have multiple wallets when we have this. I think this duplicate responses then causes a bunch of other issues in the frontend which doesn't expect this to happen (which it shouldn't).

DanThrane (changes requested):

core2/pkg/accounting/usage_report.go
```diff
@@ -120,6 +103,79 @@ func (r *internalUsageOverTimeDeltaDataPoint) ToApi() accapi.UsageReportDeltaDat
 	}
 }
 
+type UsageReportChild struct {
+	Key    string
+	Wallet AccWalletId
+}
+
+func walletToUsageReportChild(child AccWalletId) (UsageReportChild, bool) {
```
Tiggles, line 111: I haven't checked the code for any code logic issues, but I think some argument/member names are confusing:

It's not clear what kind of child it is from the name, as it's in a wallet-context, but is and ID that's being passed aorund.

`UsageReportChild.Wallet` is an ID, but we already have a type/struct called `Wallet` that is different, overloading the names.

core2/pkg/accounting/usage_report.go
```diff
@@ -120,17 +103,95 @@ func (r *internalUsageOverTimeDeltaDataPoint) ToApi() accapi.UsageReportDeltaDat
...
+	}
+}
+
 type internalUsageOverTimeAbsoluteDataPoint struct {
-	Timestamp             time.Time
-	Usage                 int64
-	UtilizationPercent100 float64
+	Timestamp time.Time
+	Usage     int64
+	Quota     int64
```
DanThrane, line 182: Breaking change. The existing charts will show 0% utilization. We should have a fallback available for this. I don't think we want to purposefully break this when we have somewhat useful data available to us.

core2/pkg/accounting/accounting_usage_generator.go
```diff
@@ -119,109 +339,157 @@ func UsageGenGenerate(api UsageGenApi, cfg accapi.UsageGenConfig) *UsageGenProje
...
+		}
+
 		minuteStep := g.Cfg.ReportingInterval
 		if minuteStep == 0 {
 			minuteStep = 5
 		}
 
+		checkpointInterval := cfg.CheckpointInterval
+		if checkpointInterval == 0 {
+			checkpointInterval = 60 // whatever default makes sense
```
DanThrane, line 422: Should be used in line 495, but it isn't.

core2/pkg/accounting/usage_report_test.go
```diff
@@ -85,6 +85,11 @@ func TestUsageGenSimulation(t *testing.T) {
 		w := e.Wallet(owner, timeAtEnd)
 		dashboards := usageRetrieveHistoricReports(e.Tm(0), timeAtEnd, w)
 
+		reportsWithProduct := make([]internalUsageReportWithProduct, len(dashboards))
```
DanThrane, line 88: If you want to pre-allocate the space, but still use append then you should set the capacity not the length. Your invocation was setting the length.

```suggestion
		reportsWithProduct := make([]internalUsageReportWithProduct, 0, len(dashboards))
```

DanThrane (changes requested):
A few minor things. I think it is important that we have a fallback for the older reports, otherwise we are just breaking a lot of usage data for no good reason. The rest looks fine.

PR #5750, opened 2026-08-26
Fixes #5744. Also updates node and postgres in launcher

### LINKED ISSUE #5744: SSHKeys allow duplicate title
Image

It seems to use the `id` on sorting/deletion, so actions are done correctly, but the user has no way of knowing which is which.

hschu12:
Changed the prefix check to verify that it is actually a valid key type.

The previous code only checked the prefix, which meant something like `ssh-rsakdfskldfjasldkfjskle` was accepted as a valid key type because it started with `ssh-rsa`.

I also added a check to ensure that a key is present after the prefix, so empty keys are rejected.

One question: should we also check whether the key has already been uploaded and inform the user that it already exists? Or is there a good reason to allow multiple SSH keys with the same key?

### DIFF
--- a/core2/pkg/migrations/00_module.go
+++ b/core2/pkg/migrations/00_module.go
@@ -38,4 +38,5 @@ func Init() {
 	db.AddMigration(grantV3())
 	db.AddMigration(grantV4())
 	db.AddMigration(projectsV5())
+	db.AddMigration(sshV1())
 }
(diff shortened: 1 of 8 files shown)


DanThrane (changes requested):
Needs to be done in a single transaction if we want to truly fix the issue. Ideally use make them unique in (user, lower(title)). Would need a db migration that also renames any duplicates we might have.

DanThrane (approved):

PR #5751, opened 2026-08-27
Changes to visual of your allocations

### LINKED ISSUE #5261: When product type/category does not contain any active allocations it shows nothing
When you do not have any active allcoations in a product type/category it shows nothing on the allocation page.
This leads to confusion. Instead of realizing that they have run out of resources due to time limitation users believe that they have just disappeared. 

On the image below the project has had compute resources and will also be granted some in one month, but expanding the compute tab just shows empty. The storeage however shows all since it has a currently active allocation

I would say either show all allocations or maybe shown all allocations that where "active" a month ago and forward.

Image

hschu12:
Fixes #5261 

Given following allocations
- CPU
  -  100 CH  active
  -  1000 CH expired 10 months ago
  -  10 CH expired 1 month ago
- Storage
  - 10 GB Future activation
  - 100 GB active
- Public IP
  - 1 IP expired 1 month ago
- License 
  - 1 Licens Future activation 

With the implemented changes:
- Allocations expired later than 3 month before current time will not be shown
- Any allocation newer than this 3 month will be shown, including future
- Wallets/products without allocations matching the rules above will be removed

Before changes:
Screenshot 2026-08-27 at 15 08 29

After changes: 
Screenshot 2026-08-27 at 15 23 06

### DIFF
--- a/frontend-web/webclient/app/Accounting/index.ts
+++ b/frontend-web/webclient/app/Accounting/index.ts
@@ -820,6 +820,27 @@ export function normalizePeriodForComparison(period: Period): Period {
     return {start: Math.floor(period.start / 1000) * 1000, end: Math.floor(period.end / 1000) * 1000};
 }
 
+export function allocationIsVisible(
+    alloc: Allocation,
+    now: number,
+): boolean {
+    const threeMonthsAgo = new Date(now);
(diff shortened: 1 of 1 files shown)


DanThrane (approved):
Better than what we have and a simple enough solution.

PR #5761, opened 2026-09-08
Fixes #5760 

### LINKED ISSUE #5760: [Mail] PIs and Admins of grant giver is recieving wrong mail
Jonas W. has recieved the approved_to_applicant mail for a project he is not a part of with the "Dear $User " is containing the wrong username.

Reason: When applications are approved the GrantEvent code is attempting to pull out PIs and Admins of the grant giver project followed by addng the receiver to the list of users to be contacted. It then uses the approved_to_applicant mail template for all. This is wrong.

If the admins and PI of the grant giver still should receive a mail, we need to reintroduce the correct mail template which seems to have disappeared, but the code still referneces it some places.

hschu12:
Fixes #5760 as well as other bugs regarding transfers.
Testes with multiple projects with different users within the project. The correct people are now getting the mail intented to them.

### DIFF
--- a/core2/pkg/accounting/grants.go
+++ b/core2/pkg/accounting/grants.go
@@ -976,6 +976,12 @@ func GrantsSubmitRevisionEx(actor rpc.Actor, req accapi.GrantsSubmitRevisionRequ
 	return int64(id), nil
 }
 
+type GrantTransferAdditionInformation struct {
+	ReceiverProject       string
+	SenderProject         string
+	ReceiverProjectAdmins []string
+}
+
 func GrantsTransfer(actor rpc.Actor, req accapi.GrantsTransferRequest) *util.HttpError {
 	now := time.Now()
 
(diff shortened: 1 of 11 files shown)


core2/pkg/accounting/grants.go
```diff
@@ -2346,19 +2402,29 @@ type grantEventType int
...
 	Actor                  rpc.Actor
 	Application            accapi.GrantApplication
+	AdditionalInfo         any
 }
 
+const (
+	NOTIFICATION_GRANT_APPLICATION_UPDATED  = "GRANT_APPLICATION_UPDATED"
+	NOTIFICATION_GRANT_APPLICATION_RESPONSE = "GRANT_APPLICATION_RESPONSE"
+	NOTIFICATION_GRANT_NEW_APPLICATION      = "GRANT_NEW_APPLICATION"
+	NOTIFICATION_GRANT_NEW_COMMENT          = "GRANT_NEW_COMMENT"
```
DanThrane: Go normally doesn't use a different naming convention for constants. So it would be `NotificationGrantApplicationupdated` and so on.

DanThrane (changes requested):
Nitpick. Rejecting only to make sure that you see it. I am okay with the rest.

DanThrane (approved):

